PR #112, opened 2026-04-29
Session containment hardening (PR1) — close setsid/at/cron evasion channels

guimard:
## Summary

Closes the SSH session evasion channels identified in our threat-model review:
- `setsid`/`nohup` detached processes that survive recorder kill
- `at` / `cron` deferred jobs running outside the recorded session
- Lingering systemd user managers running `systemd-run --user --on-active=…`

All fixes are **system configuration only** — no new setuid/setgid binary, in line with project policy.

## What it does

`ob-bastion-setup` gains a new opt-in `setup_hardening` step (skippable via `--skip-hardening`) that deploys four templates from `/usr/share/open-bastion/hardening/` into `/etc/`:

| Template | Destination | Effect |
|---|---|---|
| `logind.conf.d/open-bastion.conf` | `/etc/systemd/logind.conf.d/` | `KillUserProcesses=yes` — reaps setsid/nohup orphans at logout |
| `security/limits.d/open-bastion.conf` | `/etc/security/limits.d/` | `nproc` cap (root + `@ob-service` exempt) |
| `at.allow` | `/etc/at.allow` | Empty whitelist — root only |
| `cron.allow` | `/etc/cron.allow` | `root` only (cron itself stays up for `cron.d/open-bastion-krl`) |

`atd.service` is stopped and masked. `systemctl reload systemd-logind` (SIGHUP, non-disruptive) applies the new logind config without killing existing sessions.

A pre-flight check refuses the hardening if any non-root user has `Linger=yes` (would bypass via `systemd-run --user`). Operator must `loginctl disable-linger <user>` and re-run.

`at.allow` / `cron.allow` are never overwritten when an admin-managed file is already present; a `[WARN]` is emitted instead. A specific warning fires if an existing `cron.allow` lacks `root` (would silently break the KRL refresh job).

## Threat model coverage

| Evasion channel | Mitigation |
|---|---|
| `setsid bash …; pkill script` | `KillUserProcesses=yes` reaps cgroup at session end |
| `nohup bash &` | same |
| `echo cmd \| at now+1h` | `/etc/at.allow` empty + `atd` masked |
| `crontab -e` | `/etc/cron.allow` root-only |
| `systemd-run --user --on-active=10min` | linger pre-flight refuses hardening if any non-root has linger |
| Fork bomb | `nproc` cap (defense in depth) |

PR2 (auditd primary trace) will follow separately for traceability of execve/connect calls.

## Tests

- `tests/test_ob_bastion_setup_hardening.sh`: **20/20 PASS** (covers template content, dry-run, idempotency, linger gate, allow-list non-overwrite, `cmp -s` identical branch, regression-test against re-introducing `reload-or-restart`)
- `tests/test_ob_bastion_setup.sh`: **15/15 PASS** (no regression)
- `shellcheck`: clean
- `cmake -B build-test -S .`: configures cleanly

## Manual acceptance test

```
ssh user@bastion
setsid nohup sleep 3600 &
exit
# from root on the bastion:
ps -u user | grep sleep    # must be empty
```

## Test plan

- [ ] CI passes
- [ ] On a Debian 12 test bastion: `ob-bastion-setup --portal …`, then SSH as test user, run `setsid nohup sleep 3600 &`, disconnect, verify process is killed
- [ ] On the same host: as test user, `at now + 1 minute` should be denied
- [ ] On the same host: as test user, `crontab -e` should be denied
- [ ] As admin SSH-connected during `setup_hardening`: confirm session is **not** killed (SIGHUP only)
- [ ] Test linger gate: `loginctl enable-linger testuser` then run `setup_hardening` — must refuse with detailed error
- [ ] On a RHEL 9 / Rocky 9 box: same scenarios
- [ ] Verify Mode E KRL refresh still works after hardening (`systemctl status cron`, check `/etc/cron.d/open-bastion-krl` runs)

## Out of scope

- Pre-existing crontabs in `/var/spool/cron/crontabs/` (B4 from architect review) — accepted gap, documented as known limitation
- auditd integration (PR2)
- TOCTOU on linger (a user enabling linger between check and write) — documented follow-up

## Architect review

REJECTED then APPROVED after fixing 3 blockers (B1 conffile-doc inconsistency, B2 reload-or-restart kills sessions, B3 missing linger pre-check) + 3 nitpicks.

## Security review

Clean. Zero findings at confidence ≥ 0.8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build/
 build-core/
 build-desktop/
+build-test/
 
 # Editor files
 *~
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,6 +226,22 @@ install(PROGRAMS ${CMAKE_SOURCE_DIR}/scripts/ob-ssh-cert
     DESTINATION ${CMAKE_INSTALL_BINDIR}
 )
 
+# Install hardening templates (session containment - PR1)
+# These are deployed by ob-bastion-setup into /etc/{systemd/logind.conf.d,security/limits.d,at.allow,cron.allow}
+# See doc/hardening.md for rationale.
+install(FILES ${CMAKE_SOURCE_DIR}/config/hardening/logind.conf.d/open-bastion.conf
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/hardening/logind.conf.d
+)
+install(FILES ${CMAKE_SOURCE_DIR}/config/hardening/security/limits.d/open-bastion.conf
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/hardening/security/limits.d
+)
+install(FILES ${CMAKE_SOURCE_DIR}/config/hardening/at.allow
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/hardening
+)
+install(FILES ${CMAKE_SOURCE_DIR}/config/hardening/cron.allow
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/hardening
+)
+
 # Install setup scripts (admin-facing)
 install(PROGRAMS ${CMAKE_SOURCE_DIR}/scripts/ob-bastion-setup
     DESTINATION ${CMAKE_INSTALL_SBINDIR}
--- a/config/hardening/at.allow
+++ b/config/hardening/at.allow
@@ -1,0 +1,3 @@
+# Open Bastion: only users listed here can use at(1).
+# Empty (root only) by design — non-sudo users must not schedule
+# deferred commands that would run outside the recorded session.
--- a/config/hardening/cron.allow
+++ b/config/hardening/cron.allow
@@ -1,0 +1,4 @@
+# Open Bastion: only users listed here can use crontab(1).
+# Add admin/sudoer accounts as needed. The package needs root to
+# manage system crontabs in /etc/cron.d/.
+root
--- a/config/hardening/logind.conf.d/open-bastion.conf
+++ b/config/hardening/logind.conf.d/open-bastion.conf
@@ -1,0 +1,5 @@
+# Open Bastion: kill all user processes at logout, including those
+# detached via setsid/nohup. Required for session containment so users
+# cannot leave background processes running after disconnect.
+[Login]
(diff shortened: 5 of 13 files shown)


guimard:
Addressed in 29d81ed. Recap of how each Copilot comment is resolved:

**Comment 1 — test header inaccurate** (`tests/test_ob_bastion_setup_hardening.sh:3-4`)
Updated the header to describe the actual sandbox model (sandbox + DRY_RUN toggling + stubbed privileged ops), since several tests genuinely run with `DRY_RUN=false`.

**Comment 2 — `install_hardening_file` failures swallowed by `|| true`**
Replaced each `|| true` with an explicit `if ! install_hardening_file …; then warn …; HARDENING_RESULT=partial; fi`. Each sub-step (logind, limits, at.allow, cron.allow, mask atd, reload logind) now downgrades the result to `partial` on failure and emits a specific `[WARN]` line, so the run no longer claims success when a critical template was not installed. We did not promote a failure to fatal because the rest of the step still has value (e.g. logind failed but limits.d went through), and the operator now sees the partial state both in the closing line and in the summary.

**Comment 3 — final "Hardening applied" line is misleading**
Replaced the unconditional `info "Hardening applied …"` with a `case "$HARDENING_RESULT" in …` that prints one of:
- `applied` → `info "Hardening applied. Verify with: loginctl show-user <user> | grep Linger"`
- `partial` → `warn "Hardening applied with warnings — review the output above before relying on containment"`
- `dry-run` → `info "[DRY-RUN] Hardening would be applied. Re-run without --dry-run to commit."`

**Comment 4 — `print_summary` reads `ENABLE_HARDENING` only**
`print_summary` now reads the new global `HARDENING_RESULT` and prints distinct lines for each state (`applied` / `partial` / `dry-run` / `declined` / `failed` / `not-applied`). Concretely, a host where the operator declined the confirm prompt now shows:
```
  ✗ Session containment hardening (declined at confirm prompt)
```
instead of the previous (incorrect) checkmark.

The new `HARDENING_RESULT` global is documented inline at the top of the script with its full state machine.

Tests: `tests/test_ob_bastion_setup_hardening.sh` 20/20 PASS, `tests/test_ob_bastion_setup.sh` 15/15 PASS, shellcheck clean.

Side note: also added `build-test/` to `.gitignore` so local CMake artefacts stop sneaking into commits.

PR #113, opened 2026-04-29
Primary audit trace via auditd (PR2) — opt-in syscall-level traceability

guimard:
## Summary

Adds an **opt-in** primary audit trail via `auditd`, complementing PR1 (#112) containment. Where session recording captures pty I/O (replay convenience), auditd captures syscalls (tamper-resistant evidence) — independent of pty, shell, or session.

Activate with:
```
ob-bastion-setup --portal https://… --enable-audit-trace
```

Default: **off**. `auditd` is a `Recommends:` (Debian) / `Recommends:` (RPM 4.13+) — never installed silently.

## What it does

`setup_audit_trace` deploys two artefacts (templates from `/usr/share/open-bastion/audit/`):

| Destination | Mode | Content |
|---|---|---|
| `/etc/audit/rules.d/open-bastion.rules` | 0640 root:root | execve+execveat / connect rules (auid≥1000), watches on /etc/passwd, /etc/shadow, /etc/group, /etc/gshadow, /etc/sudoers(.d), /etc/ssh/sshd_config(.d), /var/lib/open-bastion/sessions/, /etc/open-bastion/ |
| `/etc/cron.daily/open-bastion-audit-rotate` | 0755 root:root | SIGUSR1 to auditd daily — combined with `num_logs=7` gives ~1 week local retention |

Then `augenrules --load` and `systemctl restart auditd`. Restart is non-disruptive (auditd is not session-bound, unlike logind in PR1).

**`/etc/audit/auditd.conf` is intentionally NOT modified.** It is an admin-tunable file owned by the `audit` package; we use the drop-in mechanism (`rules.d/`) where it exists, and document the recommended retention values (`num_logs=7`, `max_log_file=50`, `max_log_file_action=ROTATE`) for the admin to apply manually.

If `auditd` is not installed, `setup_audit_trace` warns and skips — the rest of `ob-bastion-setup` continues normally.

## Threat model coverage

| Channel | Covered |
|---|---|
| `execve` by non-system user | ✓ |
| `execveat` (syscall #322) | ✓ — added after security review |
| TCP/UDP `connect()` | ✓ |
| Sensitive file modifications (passwd, shadow, sudoers, sshd) | ✓ |
| Tampering with own session recording | ✓ (watch on `/var/lib/open-bastion/sessions/`) |
| `sendto`/`sendmsg` on unconnected UDP socket (DNS-tunnel exfil) | ✗ documented, opt-in extension |
| `io_uring`-based clients | ✗ documented, opt-in extension |

## Tests

- `tests/test_ob_bastion_setup_audit.sh`: **11/11 PASS** (default off, --enable flag, refuses without templates, warns when auditd absent, no auditd.conf modification, regression test against re-introducing the `audit_set_conf_key` helper, content of rules and cron templates, main() gating, summary line)
- `tests/test_ob_bastion_setup.sh`: **15/15 PASS** (no regression)
- `shellcheck`: clean
- `cmake -B build-test -S .`: configures and builds cleanly

## Reviews

- **Architect**: APPROVED — no blockers, 5 nitpicks documented, 5 follow-ups noted (audisp-syslog, --disable-audit-trace, etc.)
- **Security**: 2 findings, both fixed in commit 33ae2b0:
  - MEDIUM: `execveat` evasion → `-S execveat` added to both arch rules
  - LOW: network rule scope → inline comment + doc Limitations clarified

## Test plan

- [x] CI passes
- [ ] On a Debian 12 test bastion with `auditd` installed: `ob-bastion-setup --portal … --enable-audit-trace`, verify `auditctl -l` lists `ob-exec` and `ob-connect`, run `ls /tmp; sleep 1; ausearch -k ob-exec -ts recent` and confirm execve events
- [ ] On the same host: try `setsid bash -c 'ls; exit'` and confirm `ausearch -k ob-exec` captures both `setsid` and `bash` execs
- [ ] On the same host: write a 5-line C program that calls `syscall(SYS_execveat, …)`, run it, confirm `ausearch -k ob-exec` captures the resulting child
- [ ] On the same host without auditd installed: `ob-bastion-setup --enable-audit-trace` must WARN and continue (not fail)
- [ ] Verify `cron.daily` entry rotates audit.log after 24h+
- [ ] On RHEL 9 / Rocky 9: same scenarios
- [ ] Verify no setuid/setgid binary added to package: `find debian/tmp -perm /6000 -ls`

## Out of scope (follow-ups)

- `audisp-syslog` / journal forwarding to a remote collector (best practice for tamper-evident off-host audit)
- `--disable-audit-trace` symmetric flag
- Adding `execveat` is done; adding `sendto`/`sendmsg` is documented as opt-in extension (too chatty by default)
- Locked rules `-e 2` (currently commented in template) — production hardening, separate decision

## Relation to PR1

PR1 (#112) closes containment evasion (setsid/at/cron). PR2 adds traceability.

Both are opt-in flags on `ob-bastion-setup`:
- `--enable-hardening` (PR1) — system-wide containment
- `--enable-audit-trace` (PR2) — syscall trace

They can be combined: `ob-bastion-setup --portal … --enable-hardening --enable-audit-trace`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,6 +231,17 @@ install(PROGRAMS ${CMAKE_SOURCE_DIR}/scripts/ob-bastion-setup
     DESTINATION ${CMAKE_INSTALL_SBINDIR}
 )
 
+# Install auditd templates (consumed by `ob-bastion-setup --enable-audit-trace`).
+# Read-only templates under /usr/share/open-bastion/audit/. The setup function
+# copies them into /etc/audit/rules.d/ and /etc/cron.daily/ at deployment time.
+# These are NOT dpkg conffiles / rpm config(noreplace) — see doc/audit.md.
+install(FILES ${CMAKE_SOURCE_DIR}/config/audit/rules.d/open-bastion.rules
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/audit/rules.d
+)
+install(PROGRAMS ${CMAKE_SOURCE_DIR}/config/audit/cron.daily/open-bastion-audit-rotate
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/open-bastion/audit/cron.daily
+)
+
 install(PROGRAMS ${CMAKE_SOURCE_DIR}/scripts/ob-backend-setup
     DESTINATION ${CMAKE_INSTALL_SBINDIR}
 )
--- a/config/audit/cron.daily/open-bastion-audit-rotate
+++ b/config/audit/cron.daily/open-bastion-audit-rotate
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Open Bastion: rotate auditd logs daily so that with num_logs=7
+# (configured in /etc/audit/auditd.conf) we keep ~1 week of history.
+# Installed by `ob-bastion-setup --enable-audit-trace`.
+
+# Trigger auditd rotation (same as `service auditd rotate`).
+# Old auditd: SIGUSR1 to the daemon. Modern: systemctl kill -s USR1.
+if command -v systemctl >/dev/null 2>&1 && systemctl is-active auditd >/dev/null 2>&1; then
+    systemctl kill -s USR1 auditd 2>/dev/null || true
+elif [ -r /var/run/auditd.pid ]; then
+    kill -USR1 "$(cat /var/run/auditd.pid)" 2>/dev/null || true
+fi
+
+exit 0
--- a/config/audit/rules.d/open-bastion.rules
+++ b/config/audit/rules.d/open-bastion.rules
@@ -9,10 +9,16 @@
 -f 1
 
 # === EXECVE: every command run by a logged-in non-system user ===
--a always,exit -F arch=b64 -S execve -F auid>=1000 -F auid!=unset -k ob-exec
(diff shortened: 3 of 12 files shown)


guimard:
Addressed in 63c3b58. Recap:

**Comment 1 — unguarded `install` calls** (`scripts/ob-bastion-setup:731,736`)
Real bug. With `set -euo pipefail`, an `install` failure (missing `/etc/audit/rules.d/`, RO filesystem, perms-denied) would abort the entire `ob-bastion-setup` run, never reaching `main()`'s `setup_audit_trace || warn …` recovery path. Both calls now wrapped in `if ! install …; then error; return 1; fi` so failures surface as a clean non-zero return from `setup_audit_trace`.

**Comment 2 — trailing comma in `Recommends: auditd,`** (`debian/control:25`)
Removed. `Recommends: auditd` (no comma) is consistent with `Recommends: fonts-noto` elsewhere in the same file. `auditd` is the only recommendation for the `open-bastion` package, so the comma is gratuitous.

**Comment 3 — Test 8 missing in numbering** (`tests/test_ob_bastion_setup_audit.sh:258+`)
Cosmetic but real. The numbering gap was a leftover from commit `991556b` which removed the three `audit_set_conf_key` tests (originally tests 6/7/8). Renumbered the remaining headers: 9→8, 10→9, 11→10, 12→11.

Validations: `bash -n` clean, `shellcheck` clean, 11/11 audit tests pass, 15/15 regression tests pass.

PR #124, opened 2026-06-13
fix(setup): sshd drop-in must outrank cloud-init (password auth stayed enabled)

guimard:
## Problem (found during lab validation)

`sshd -T` on a freshly-deployed bastion **and** backends showed `passwordauthentication yes` — the cert-only lockdown (R-S1) was silently defeated. A plain `ssh user@host` without a cert fell back to a **password prompt** instead of being refused.

Root cause: cloud images ship `/etc/ssh/sshd_config.d/50-cloud-init.conf` containing `PasswordAuthentication yes`. sshd uses the **first** value seen for keywords like `PasswordAuthentication` / `PermitRootLogin` / `KbdInteractiveAuthentication`, and the `Include` directory is read alphabetically. The setup scripts wrote `50-open-bastion-<role>.conf`, which sorts **after** `50-cloud-init.conf` (`c` < `o`), so cloud-init's `yes` won and our `no` was ignored.

## Fix

Write the drop-in as **`00-open-bastion-<role>.conf`** so it is read before any distro drop-in and the open-bastion hardening (`PasswordAuthentication no`, `PermitRootLogin no`, `KbdInteractiveAuthentication no`, …) wins. The legacy `50-` file is removed on re-run.

## Regression test — `tests/test_sshd_dropin_precedence.sh`

- Extracts the drop-in name the setup scripts actually write (so a future rename back to `50-` fails the test) and asserts it sorts before `50-cloud-init.conf`.
- In docker, with both drop-ins present, a real `sshd -T` confirms `passwordauthentication=no`; a negative control with the legacy `50-` name yields `yes`, proving the ordering is the cause.

## Lab validation (3 VMs, all carrying cloud-init's `50-cloud-init.conf`)

After deploy: `sshd -T` → `passwordauthentication no` on bastion + both backends; the `50-` drop-in is gone (replaced by `00-`); the cert hop bastion→backend still works; and `ssh dwho@host` **without** a cert is now refused (`Permission denied (publickey)`) with no password fallback.

Independent of the `/var` token MR (#123); this is a pre-existing hardening gap in `ob-bastion-setup` / `ob-backend-setup`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **sshd hardening drop-in could be silently overridden by cloud-init.** Cloud
+  images ship `/etc/ssh/sshd_config.d/50-cloud-init.conf` with
+  `PasswordAuthentication yes`, and sshd keeps the *first* value seen while
+  `Include` expands the drop-in directory alphabetically. The open-bastion
+  drop-in was written as `50-open-bastion-{bastion,backend}.conf`, which sorts
+  *after* `50-cloud-init.conf`, so password authentication stayed enabled on
+  freshly provisioned bastions and backends. `ob-bastion-setup` /
+  `ob-backend-setup` now write `00-open-bastion-{bastion,backend}.conf` (and
+  remove the legacy `50-` file on rerun) so the cert-only lockdown wins.
+
 ### Changed
 
 - **Server token relocated from `/etc/open-bastion/token` to
--- a/doc/admin-guide.md
+++ b/doc/admin-guide.md
@@ -248,7 +248,7 @@ EOF
 ### Step 6: Configure SSH with Recording
 
 ```bash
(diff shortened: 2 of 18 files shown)


guimard:
Addressed the Copilot review (commit `8ef5e7f`):

1. **CI glob** — renamed the test to `tests/test_ob_sshd_dropin_precedence.sh` so the unit-test job (`tests/test_ob_*.sh`) actually runs it.
2. **Silent extraction failure** — fail fast if the drop-in name cannot be extracted from the setup scripts (an empty string sorted first and would have masked a regression).
3. **Weak negative control** — the docker control now *asserts* that the legacy `50-` name resolves to `passwordauthentication=yes` (was a NOTE-only), so the test really proves ordering is the root cause.
4 & 5. **Stale `50-` references** — swept the remaining `50-open-bastion-{bastion,backend}.conf` references to `00-` across `doc/admin-guide.md`, `doc/bastion-architecture.md`, the docker demos (`cert`/`maxsec`/`token`/`token-svc`) and the integration tests, so copy/paste onto cloud images cannot reintroduce the cloud-init race. The only remaining `50-open-bastion-backend.conf` is the intentional negative control in the precedence test, and the historical rename note in `CHANGELOG.md`.

Test passes locally including the docker dynamic check with the now-asserting negative control.

PR #126, opened 2026-06-13
feat(bastion): rename ob-ssh-proxy → ob-ssh, add ob-scp, fix interactive TTY

guimard:
## Summary

Renames the bastion-to-backend connector **`ob-ssh-proxy` → `ob-ssh`**, adds a new **`ob-scp`**, factors the shared cert-vouching logic into a sourced library **`ob-cert-lib.sh`**, and fixes the interactive TTY behaviour of the connector. No back-compat symlink (new project).

### Added
- **`ob-scp`** — file copy bastion↔backend and backend↔backend with a vouched certificate, forced through the bastion (`scp -3`) so the connection's source address matches the cert's pinned address. All remote endpoints must share the same remote user (one vouched cert = one principal).

### Changed
- **`ob-ssh-proxy` → `ob-ssh`**; shared minting/config/token logic extracted into `scripts/ob-cert-lib.sh` (installed at `/usr/lib/open-bastion/ob-cert-lib.sh`).
- Packaging (CMake/debian/rpm), man pages (new `ob-ssh.1` describing **cert vouching** instead of the dead JWT mechanism, new `ob-scp.1`), docs, demos and tests updated.

### Fixed
- **Interactive `ob-ssh`: double echo + Ctrl-C / a failing command tore down the connection.** The connector relied on ssh's TTY auto-detection across the bastion-pty → backend-pty hop, which could leave the bastion-side tty in cooked mode (double echo) and deliver signals to the connector instead of the remote shell. `ob-ssh` now sets `-tt` when stdin is a terminal (bastion-side tty goes raw → single echo, signals/failures act on the remote shell) and `-T` otherwise.

## Tests
- New `tests/test_ob_scp.sh` + reworked `tests/test_ob_ssh.sh` (function tests now source `ob-cert-lib.sh`); both pass, shellcheck clean.
- **e2e on the 3-VM lab**: `ob-ssh` → single echo, survives a failing command, clean logout; `ob-scp` bastion→backend ✓ and backend→backend via `scp -3` ✓.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


guimard:
Addressed the Copilot review (commit \`96e8338\`):

**Real bugs**
- \`ob-cert-lib.sh\` (mint_ephemeral_cert): the **ssh-keygen** and **request_bastion_cert** failure paths now \`rm -rf\` the tmpfs directory before returning. The caller only arms its cleanup trap after minting succeeds, so these early returns previously left the ephemeral private key on disk. Fixed both.

**Consistency**
- \`ob-ssh\` usage: removed \`SERVER_GROUP\` from the config example (the library accepts-but-ignores it; the man page no longer documents it).
- \`doc/design/bastion-cert-vouching.md\`: the "why the JWT mechanism is dead" paragraph now attributes it to *the original connector (then named ob-ssh-proxy)* instead of the new cert-based \`ob-ssh\`.

**Demo docs/comments (JWT → certificate vouching)**
- \`docker-demo-{cert,token,token-svc}\`: updated the \`ssh-proxy.conf\` comments, README flow descriptions, mermaid diagrams, "Bastion JWT Verification" sections and the access-control Notes to describe \`ob-ssh\` minting an ephemeral certificate via \`/pam/bastion-cert\` and the backend validating it — instead of the dead \`LLNG_BASTION_JWT\` / \`/pam/bastion-token\` transport.

The two low-confidence notes on \`tests/test_ob_ssh.sh\` (the config-permissions mock returns a non-root owner, so the ownership check short-circuits before the group-writable branch) are a pre-existing inaccuracy inherited from the old test; left as-is for this PR since the assertions still pass and it is orthogonal to the rename. Happy to tighten it in a follow-up if you prefer.

CI green (ShellCheck scans \`./scripts\`, which stays clean; the LOGIN_RESP/\`i\` SC2034 warnings are pre-existing and only in docker-demo entrypoints, which CI does not scan).

PR #140, opened 2026-06-15
fix: robust Ansible Mode E deploy + fail enrollment without refresh token

guimard:
Three fixes found while deploying a Mode E bastion through the ob-builder Ansible artifact.

## 1. `ob-enroll`: fail when offline requested but no refresh token issued
When the scope includes `offline_access` but the portal returns **no** `refresh_token`, `save_token` previously wrote the (unrenewable) token and returned 0. NSS/SSO then worked for ~1h, and `ob-bastion-setup` only aborted later — leaving a dead token that got reused on the next run.

Now: remove the token and `exit 1` with an actionable message (enable `AllowOffline`, allow `offline_access`, deploy `oidc-device-organization >= 0.3.3` which keeps `offline_access` instead of stripping it). A refresh-less token is still accepted when `offline_access` was **not** requested.

> Context: a portal running an old `oidc-device-organization` stripped `offline_access` from the device-code scope, so the device grant returned no refresh token while the auth-code grant did. `ob-bastion-setup` (Mode E) rightly refused; this change surfaces it at enrollment instead.

## 2. Ansible role: gate the `Restart sshd` handler on `ob_auto_setup`
With `ob_auto_setup: true`, `ob-*-setup` already restarts sshd, **and** Mode E locks `sudo` behind an LLNG token — so a `become` handler flushed afterwards fails with "Missing sudo password" on an otherwise-successful deploy. The handler now runs only when `ob_auto_setup` is false (Ansible-managed config without the setup script).

## 3. `inventory.yml.example`: document SSH connection variables
Commented `ansible_host` / `ansible_user` / `ansible_ssh_private_key_file` / `ansible_become` on the first host, plus the `IdentitiesOnly=yes` tip for "Too many authentication failures" and an `-m ping` connectivity check.

## Tests
- `tests/test_ob_enroll.sh`: +2 cases (no-refresh ⇒ fail+remove; no-offline ⇒ ok). Full suite **27/27**.
- Templates validated as YAML; `bash -n scripts/ob-enroll` clean.

(2) and (3) are ob-builder template changes; regenerate artifacts to pick them up.

### DIFF
--- a/admin-builder/templates/ansible/inventory.yml.example.in
+++ b/admin-builder/templates/ansible/inventory.yml.example.in
@@ -8,6 +8,24 @@
 #
 # Override ob_client_secret per-host (or group_vars) via ansible-vault:
 #   ansible-vault encrypt_string 's3cr3t' --name ob_client_secret
+#
+# ── SSH connection variables ──────────────────────────────────────────────────
+# How Ansible reaches and escalates on each host (NOT the open-bastion users —
+# this is the admin account used only to deploy). Set per host or per group:
+#
+#   ansible_host                 IP/FQDN actually used to reach the host. The
+#                                inventory name above can stay logical/symbolic.
+#   ansible_user                 SSH login of the deploy account (needs sudo).
+#   ansible_ssh_private_key_file Path to that account's private key.
+#   ansible_become: true         Escalate to root with sudo for the tasks.
+#                                Add --ask-become-pass (-K) on the CLI, or set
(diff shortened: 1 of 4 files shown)


guimard:
Good catch — fixed: added `| bool` so `when: not (ob_auto_setup | default(true) | bool)` handles a `--extra-vars ob_auto_setup=false` string override correctly (otherwise "false" is truthy in Jinja and the handler would be wrongly skipped).

PR #169, opened 2026-06-18
feat(ob-builder): declare service keys (service accounts)

### LINKED ISSUE #167: ob-builder: support declaring service keys (service accounts)
## Summary

Let `admin-builder`/`ob-builder` configure **service keys** (service-account SSH
keys) as part of the generated deployment artifacts (shell installer + Ansible
role), so admins can declare accounts like `ansible`, `backup`, CI/CD at build
time instead of hand-editing `service-accounts.conf` on every server.

## Background

The on-server side already exists and needs **no C changes**:
- Service accounts are parsed from `/etc/open-bastion/service-accounts.conf`
  (INI-style: `[name]` + `key_fingerprint`, `sudo_allowed`, `sudo_nopasswd`,
  `gecos`, `shell`, `home`, `uid`, `gid`) — see `src/service_account.c` and
  `doc/service-accounts.md`.
- `ob-builder` already generates `openbastion.conf` and deploys it for shell and
  Ansible targets; it does **not** currently emit `service-accounts.conf` nor a
  `service_accounts_file` directive.

## Scope

**`admin-builder/ob-builder`**
- New questionnaire step (optional, applies to bastion/standalone/backend):
  "define service accounts?" → loop collecting per account:
  username, SSH key fingerprint (`SHA256:…`), `sudo_allowed`, `sudo_nopasswd`,
  shell, home, optional gecos.
- New YAML config schema for non-interactive mode:
  ```yaml
  enable_service_accounts: yes
  service_accounts:
    - name: ansible
      key_fingerprint: "SHA256:..."
      sudo_allowed: true
      sudo_nopasswd: true
      shell: /bin/bash
      home: /var/lib/ansible
    - name: backup
      key_fingerprint: "SHA256:..."
      sudo_allowed: false
  ```
- Build placeholders: render `service-accounts.conf` (INI), base64 for the shell
  installer, YAML/content for Ansible defaults; set `SERVICE_ACCOUNTS_ENABLED`.

**`admin-builder/lib/`**
- `validators.sh`: `is_valid_username`, `is_valid_fingerprint`,
  `is_valid_shell`, `is_valid_home` (mirror the C-side validation rules).
- `prompts.sh`: an interactive add/edit/remove loop helper.

**Templates**
- `templates/shell/openbastion.conf.in` + `installer.sh.in`: emit the
  `service_accounts_file = …` directive and write
  `/etc/open-bastion/service-accounts.conf` (mode 0600, root) from embedded
  base64; add a `--skip-service-accounts` flag.
- `templates/ansible/role/defaults/main.yml.in`: `ob_service_accounts_enabled`
  + `ob_service_accounts_content`.
- `templates/ansible/role/templates/openbastion.conf.j2`: conditional
  `service_accounts_file` directive.
- `templates/ansible/role/tasks/main.yml.in`: task to deploy
  `service-accounts.conf` (owner root, mode 0600) when enabled.

## Design questions to settle in the PR
- Per-role vs global accounts (recommend: global, overridable per host via
  Ansible host_vars/group_vars).
- Whether to support post-deploy updates without re-running ob-builder (doc the
  manual edit path at minimum).

## Acceptance criteria
- `ob-builder` (interactive **and** YAML) can declare ≥1 service account.
- Generated shell installer writes a correct, validated `service-accounts.conf`
  (0600, root) and references it from `openbastion.conf`.
- Generated Ansible role deploys the same on target hosts.
- Invalid fingerprints/usernames/shells/homes are rejected at build time.
- `doc/service-accounts.md` updated with the ob-builder integration path.

guimard:
## Summary

`ob-builder` can now declare **service keys** (service-account SSH keys) as part
of the generated deployment artefacts, so admins declare accounts like
`ansible`, `backup` or CI/CD once at build time instead of hand-editing
`/etc/open-bastion/service-accounts.conf` on every server.

**No PAM-module change** — `src/service_account.c` already parses that file; this
PR is entirely in the builder + templates + docs.

## What's new
- **Interactive questionnaire** step (applies to every role): loops over
  name / fingerprint / sudo / shell / home / gecos, validating each field.
- **Non-interactive `--config`**: a `service_accounts:` YAML list, parsed by a
  dedicated, **yq-independent** block scanner so behaviour is identical whether
  or not `yq` is installed (the flat parsers can't read a list).
- **Build-time validation** mirroring `src/service_account.c`: username
  (`^[a-z_][a-z0-9_-]*$`), fingerprint (`SHA256:`/`MD5:` + base64 charset),
  absolute shell/home; duplicate account names rejected.
- **Shell installer** writes `/etc/open-bastion/service-accounts.conf`
  (`0600 root:root`) from an embedded base64 blob and sets
  `service_accounts_file` in `openbastion.conf`.
- **Ansible role** carries the accounts as `ob_service_accounts_content`
  (a YAML block scalar, overridable per host/group via `host_vars`/`group_vars`)
  and deploys the file when `ob_service_accounts_enabled` is true.

## Config example
```yaml
service_accounts:
  - name: ansible
    key_fingerprint: "SHA256:abc123..."   # ssh-keygen -lf key.pub
    sudo_allowed: true
    sudo_nopasswd: true
    shell: /bin/bash
    home: /var/lib/ansible
    gecos: Ansible Automation
  - name: backup
    key_fingerprint: "SHA256:xyz789..."
    sudo_allowed: false
    shell: /bin/sh
```

## Files
- `admin-builder/ob-builder`: record model, YAML block parser, interactive loop,
  validation, INI renderer, placeholder map, recap line, usage text.
- `admin-builder/lib/validators.sh`: `is_valid_service_username`,
  `is_valid_fingerprint`, `is_valid_abs_path`.
- Templates: shell `openbastion.conf.in` + `installer.sh.in` (deposit), Ansible
  `defaults/main.yml.in` + `openbastion.conf.j2` + `tasks/main.yml.in`.
- Docs: `doc/service-accounts.md` (ob-builder integration), `ob-builder(1)`,
  `admin-builder/README.md`, `CHANGELOG.md`.

## Testing
- `tests/test_ob_builder_service_accounts.sh` (picked up by CI's
  `tests/test_ob_*.sh` loop): validators, YAML block parsing (2 accounts +
  none), record validation accept/reject, INI rendering — 6/6 pass.
- End-to-end render check (offline, with a fake work dir): the shell installer
  embeds and deposits the conf and sets `service_accounts_file`; the Ansible
  `defaults/main.yml` is **valid YAML** for both enabled and disabled cases with
  **no leftover `@@…@@` placeholders**.
- `shellcheck --severity=warning` clean on `ob-builder` and `validators.sh`;
  `ob-builder(1)` lints with no `man` warnings.

Closes #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Documentation
 
+- **Service-account security model documented.** `doc/service-accounts.md` now
+  spells out that ob-builder deposits the fingerprint only, so the public key
+  must still be authorized at the SSH layer (`authorized_keys`, or the
+  `ob-service-account-keys` `AuthorizedKeysCommand` helper for Mode E); that
+  service accounts use direct key auth (no `ob-ssh`), so a native ProxyJump hop
+  through the bastion is **not session-recorded**; and that service-account sudo
+  is granted locally with **no LLNG token, even in Mode E**. New EBIOS risks
+  R-S24 (sudo without token) and R-S25 (unrecorded ProxyJump hop) in
+  `doc/security/99-risk-reduce.md`. `local-test/deploy-shell.sh` gained an
+  end-to-end service-account check on the standalone VM.
 - **Backend access guidance corrected.** The admin guide showed a
   `ProxyCommand`/`ProxyJump` snippet for reaching a backend in one command, but
   `ob-ssh` re-originates a full interactive session (it is not a `-W` stdio
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   bastion and no agent forwarding. Connects to a single endpoint
   (`[user@]backend[:path]`); options after the connector's own flags pass
   straight through to `sftp(1)`. See `ob-sftp(1)`.
-
+- **`ob-builder` can declare service accounts.** The builder now collects
+  SSH-key-only local accounts (ansible, backup, CI/CD, …) — interactively or via
+  a `service_accounts:` list in the `--config` YAML — validates each entry
+  (name, `SHA256:`/`MD5:` fingerprint, absolute shell/home) at build time, and
+  bakes them into both outputs: the shell installer writes
+  `/etc/open-bastion/service-accounts.conf` (`0600 root:root`) and the Ansible
+  role carries them as `ob_service_accounts_content` (overridable per
+  host/group). `service_accounts_file` is set in the generated
+  `openbastion.conf`. No PAM-module change — `src/service_account.c` already
+  parses that file. See `doc/service-accounts.md` and `ob-builder(1)`.
 - **Session-recording retention (`ob-session-prune`).** A new daily timer
   (`ob-session-prune.timer`, enabled at install) bounds the recordings store,
   which matters because recording is fail-closed — a full disk refuses new
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   host/group). `service_accounts_file` is set in the generated
   `openbastion.conf`. No PAM-module change — `src/service_account.c` already
   parses that file. See `doc/service-accounts.md` and `ob-builder(1)`.
+  Validated end-to-end on a Mode E VM (`local-test/deploy-shell.sh`). ob-builder
+  warns when an account would be unusable on the target: a `home`/`shell` outside
+  the approved lists (silently dropped by the PAM module) or a missing fixed
+  `uid`/`gid` (NSS cannot resolve it for sshd's pre-auth lookup, so it is
+  unreachable over SSH unless it already exists locally). `doc/service-accounts.md`
+  documents these requirements (including not reusing a system username).
 - **Session-recording retention (`ob-session-prune`).** A new daily timer
   (`ob-session-prune.timer`, enabled at install) bounds the recordings store,
   which matters because recording is fail-closed — a full disk refuses new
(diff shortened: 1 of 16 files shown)


guimard:
## Follow-up: VM test + service-account trust model (commit dcbc6b2)

Added an end-to-end service-account check to `local-test/deploy-shell.sh` and documented the trust model raised in review.

**Two questions answered (grounded in the code):**

1. **ProxyJump.** A service account = direct SSH key auth, per-server local config, independent of the bastion cert-vouching. It can `ssh` directly to any backend where it's configured. A native `ssh -J bastion backup@backend` *can* work if `backup` is configured on both sides and the bastion permits TCP forwarding — **but the bastion `ForceCommand` recorder does not cover the `direct-tcpip` channel, so that hop is NOT recorded**. Service accounts can't use `ob-ssh` (no SSO voucher), so the intended pattern is to point them **directly at targets**, not relay through the bastion. (`AllowTcpForwarding` is not disabled by the setups, so the channel is open by default.)

2. **sudo without token.** Confirmed (`pam_openbastion.c:3232‑3269`): a service account's sudo comes entirely from `service-accounts.conf` (`sudo_allowed`/`sudo_nopasswd`), granted locally with **no LLNG call, even in Mode E**. A sudo-enabled service key is a standing local privilege that escapes the SSO-gated sudo model.

**Gap surfaced:** ob-builder deposits the **fingerprint only**. That's necessary but not sufficient — `sshd` must also authorize the public key (`authorized_keys` in modes A–D, or the `ob-service-account-keys` `AuthorizedKeysCommand` helper for Mode E; the latter also avoids the account-doesn't-exist-yet chicken-and-egg). Per maintainer decision, this manual step is **documented** rather than auto-wired, and the C-side helper packaging/wiring is left as-is.

**This commit:**
- `local-test/deploy-shell.sh`: the standalone config declares a `backup` service account (fingerprint = a lab key); the harness provisions the SSH-layer authorization (helper + `service-accounts.d/backup.pub`) before setup locks port 22, then asserts `backup` logs in with its key — no SSO, no cert.
- `doc/service-accounts.md`: SSH-layer authorization, ProxyJump-not-recorded, sudo-without-token sections.
- `doc/security/99-risk-reduce.md`: EBIOS risks **R-S24** (sudo without token) and **R-S25** (unrecorded ProxyJump hop).

PR #222, opened 2026-09-05
fix(security): harden file handling and bounds checks across daemon, NSS and stores

### LINKED ISSUE #184: Cache key material: read_key_file accepts a non-root-owned key file; secret_store derives from machine-id only
Two independent defects in how cache/secret encryption keys are sourced.

## 1. `read_key_file` accepts a key file owned by a non-root uid

`src/cache_key.c:335-346`:

```c
if ((st.st_mode & 0077) != 0) {
    syslog(LOG_WARNING, "... insecure permissions %04o (should be 0600)", ...);
    if (st.st_uid != 0) { close(fd); return -1; }   /* non-root rejected ONLY here */
    /* Root-owned with loose permissions: warn but allow */
}
```

The ownership check sits **inside** the loose-permissions branch. A key file that is mode 0600 but owned by another uid therefore passes with no check at all. A root-owned 0644 key file is likewise accepted with only a warning.

## 2. `secret_store` has no key-file support

`src/secret_store.c:123-177` (`derive_key`) derives from `/etc/machine-id` plus an optional config salt, and nothing else. `/etc/machine-id` is world-readable (0444), and the salt is stored alongside the ciphertext, so disk or backup theft yields the key.

`SECURITY.md:660` claims "Cache file theft -> AES-256-GCM + machine/key binding", which overstates this: for the secret store the protection is integrity only, not confidentiality.

## Scope correction

The original analysis claimed the default case is "no key file at all". That is overstated for the offline cache: `src/offline_cache.c:113` defaults to `/etc/open-bastion/cache.key` and `scripts/ob-desktop-setup:400-405` generates it (32 random bytes, `chmod 600`) whenever `--offline` is used, warning to syslog if it is missing.

Integrity is unaffected in every case — the GCM tag is verified before use.

## Suggested fix

* Move the `st.st_uid != 0` check out of the loose-permissions branch so it always runs, and fail closed on a non-0600 key file rather than warning.
* Either add key-file support to `secret_store`, or correct `SECURITY.md` to describe machine-id derivation as integrity-and-obfuscation rather than confidentiality.

---

### LINKED ISSUE #185: ob-cert-daemon reads its config and server token without the O_NOFOLLOW/fstat discipline used elsewhere
`ob-cert-daemon` custodies the bastion's server token, yet it is the one component that skips the `O_NOFOLLOW` + `fstat`-on-fd discipline used consistently elsewhere in the codebase.

`src/ob-cert-daemon.c:111-115, 139-145, 174-177`:

```c
static void load_ini(struct ob_cfg *cfg) { FILE *f = fopen(OB_CONFIG, "r"); ... }   /* no fstat, no O_NOFOLLOW */

static void load_proxy(...) { if (stat(PROXY_CONFIG,&st)!=0) return;
    if (st.st_uid!=0 || (st.st_mode&(S_IWGRP|S_IWOTH))) return;
    FILE *f = fopen(PROXY_CONFIG,"r");   /* TOCTOU: stat then open */ }

static char *read_server_token(const char *path) { FILE *f = fopen(path,"r"); ... }  /* no O_NOFOLLOW, no perm check */
```

Compare `src/config.c:29-53` (`check_file_permissions_fd`) and `nss/libnss_openbastion.c:324-357`, which both do this correctly.

## Impact

Defence in depth rather than a directly exploitable hole. `scripts/ob-bastion-setup` sets `/etc/open-bastion` to 0711 root (:722) and `openbastion.conf` / the token file to 0600 root (:563, :579, :1511), so redirecting `portal_url` to exfiltrate the Bearer token already presumes write access to a root-only directory, i.e. a prior compromise. But this is the most sensitive path in the product and it should not be the one place the hardening pattern is missing.

## Suggested fix

Apply `check_file_permissions_fd` and `O_NOFOLLOW` to `openbastion.conf`, `ssh-proxy.conf` and the server token file; replace the `stat`-then-`fopen` in `load_proxy` with open-then-`fstat`.

---

### LINKED ISSUE #187: secret_store_get decrypts into the caller buffer without a size check, before GCM authentication
`secret_store_get()` decrypts a file into the caller's buffer without ever comparing the ciphertext length to the buffer size.

`src/secret_store.c:483-486`:

```c
if (EVP_DecryptInit_ex(ctx, EVP_aes_256_gcm(), NULL, NULL, NULL) != 1 ||
    EVP_DecryptInit_ex(ctx, NULL, NULL, store->derived_key, iv) != 1 ||
    EVP_DecryptUpdate(ctx, secret, &len, ciphertext, ciphertext_len) != 1) {
```

`ciphertext_len` comes from `st.st_size` (`:467`). `secret_size` is never compared against it — it is only used for the `explicit_bzero` at `:508`. A `.enc` file whose payload exceeds the caller's buffer overflows it, and because the GCM tag is only verified afterwards in `EVP_DecryptFinal_ex`, the overflow happens *before* authentication.

## Preconditions

Writing an oversized file into the 0700 root store directory requires root, or a caller bug. The store is opened `O_NOFOLLOW` and symlinks are rejected and unlinked (`:424-430`). So this is defence in depth, not a live vulnerability.

## Suggested fix

Reject `ciphertext_len + TAG_SIZE > secret_size` before calling `EVP_DecryptUpdate`.

---

### LINKED ISSUE #189: NSS file cache is world-readable (0644): any local account can enumerate all SSO users
The NSS module's on-disk cache is created world-readable in a world-readable directory, so any local account can enumerate every SSO user who has ever logged in.

`nss/libnss_openbastion.c:678-712`:

```c
if (mkdir(CACHE_DIR, 0755) == -1 && errno != EEXIST) { ...; return; }
FILE *f = fopen(filepath, "w");
...
fchmod(fileno(f), 0644);
```

An unprivileged local user can read name, uid, gid, gecos, home and shell for all cached SSO users. The code itself comments on the sensitivity of this data at `:1437-1442`, where it refuses to cache service accounts.

## Secondary hardening gaps

* `/var/cache/nss_llng` is never provisioned by packaging — a grep over `debian/`, `rpm/`, `systemd/`, `scripts/` and `config/` finds no reference to `nss_llng`. The directory is created lazily on first use.
* `file_cache_save` uses a plain `fopen("w")` with no `O_NOFOLLOW`, no `O_EXCL` and no `geteuid() == 0` guard.

## Note on severity (correcting the original analysis)

This was reported as a local privilege escalation via a symlink attack. **That chain does not work**, for two independent reasons:

1. `query_llng_userinfo` returns `-1` immediately when `server_token` is NULL (`nss/libnss_openbastion.c:1104-1107`), and `load_server_token` (`:324-349`) enforces `O_NOFOLLOW`, `st_uid == 0` and no group/other bits on a file that `scripts/ob-bastion-setup:1511` sets to 0600 root. A non-root lookup therefore never reaches `file_cache_save()`.
2. `mkdir("/var/cache/nss_llng")` from a non-root process fails `EACCES` on a standard root:root 0755 `/var/cache`, so an attacker can never own the directory.

What remains is the information leak plus the latent hardening gap.

## Suggested fix

* Create the cache files 0600 (or 0644 in a 0700 directory), and provision the directory root-owned from packaging or `tmpfiles.d`.
* Add `O_NOFOLLOW` and a root-only write guard to `file_cache_save`.

---

### LINKED ISSUE #196: NSS module: unchecked server-supplied GID, silent Authorization header truncation, trim() UB
Three low-severity defects in `nss/libnss_openbastion.c`, grouped.

## 1. The server-supplied GID is not range-checked

`nss/libnss_openbastion.c:1259-1264`:

```c
if (json_object_object_get_ex(json, "gid", &val) && json_object_is_type(val, json_type_int)) {
    pw->pw_gid = (gid_t)json_object_get_int(val);
} else { pw->pw_gid = g_config.default_gid; }
```

The uid **is** validated against `[min_uid, max_uid]` and `NOBODY_UID` at `:1233-1240`, and the lookup fails on violation. The gid is taken verbatim, so gid 0 or a negative int32 widening to a huge `gid_t` is served to every NSS consumer. Exploitation requires a hostile or compromised LLNG portal, which is the module's trust anchor — so this is defence in depth, but `SECURITY.md`'s "UID range enforcement" claim does not cover it.

## 2. The Authorization header is silently truncated

`nss/libnss_openbastion.c:1117-1128`:

```c
char url[512];  snprintf(url, sizeof(url), "%s/pam/userinfo", g_config.portal_url);
char auth_header[512]; snprintf(auth_header, sizeof(auth_header), "Authorization: Bearer %s", server_token);
```

Both are `snprintf`-bounded so there is no memory-safety issue, but the token is read into an 8192-byte buffer at `:364`. Any server token longer than roughly 490 characters — a JWT-shaped access token easily is — produces a truncated Bearer header and a permanent 401 loop, with no diagnostic beyond "transient failure".

## 3. `trim()` forms `str - 1` on an empty string

`nss/libnss_openbastion.c:112-119`: `char *end = str + strlen(str) - 1;` is computed unconditionally, which is undefined behaviour per C11 6.5.6/8 even without a dereference. The `end > str` guard prevents any actual read or write. The sibling implementation `src/str_utils.c:21` already has the right shape (`if (*str == '\0') return str;`).

## Suggested fix

Range-check the gid like the uid; check the `snprintf` return for the Authorization header and fail loudly on truncation; align `trim()` with `str_utils.c`.

---

guimard:
Five defects of the same family: a privileged component trusting a path, a mode, or a length it never checked.

## 1. `ob-cert-daemon` skips the file discipline used everywhere else (#185)

`src/ob-cert-daemon.c` custodies the bastion's server token but was the one component reading config and secrets with plain `fopen()`:

- `load_ini()` — `fopen(OB_CONFIG, "r")`, no `fstat`, no `O_NOFOLLOW`
- `load_proxy()` — `stat(PROXY_CONFIG, ...)` **then** `fopen(...)`: a textbook TOCTOU
- `read_server_token()` — `fopen(path, "r")`, no `O_NOFOLLOW`, no permission check at all

`read_server_token()` was the worst of the three: the path comes from `SERVER_TOKEN_FILE` in `ssh-proxy.conf`, which `ob-bastion-setup` deliberately writes **0644 world-readable**. The path is attacker-visible, and the file it names is the credential the daemon presents to LLNG.

Fix: a local `open_checked()` helper modelled on `check_file_permissions_fd()` (`src/config.c:29`) and `load_server_token()` (`nss/libnss_openbastion.c:324`) — `O_NOFOLLOW`, `fstat()` on the *opened* fd, `S_ISREG`, root ownership.

Two policies, deliberately:

| File | Policy | Why |
|---|---|---|
| `openbastion.conf`, server token | strict — no group/other bits | every installer writes them `0600 root` (`debian/open-bastion.postinst:205,240`, `scripts/ob-bastion-setup:563,1511`), and `src/config.c` already enforces exactly this on the same conf for the PAM module |
| `ssh-proxy.conf` | root-owned, not group/world-**writable** | it is `0644` **on purpose** (`scripts/ob-bastion-setup:717`) — ob-ssh runs as the connecting user and reads it; it holds no secret |

**Deviation worth flagging:** applying the strict `check_file_permissions_fd` policy to `ssh-proxy.conf`, as a literal reading of the issue would suggest, would reject a file the product intentionally ships world-readable and silently drop `PORTAL_URL`/`SERVER_TOKEN_FILE`/`VERIFY_SSL` overrides. The existing writability-only policy is the correct one there; only the TOCTOU shape changed.

## 2. `read_key_file` accepts a non-root-owned key file (#184)

`src/cache_key.c` had the ownership test nested **inside** the loose-permissions branch:

```c
if ((st.st_mode & 0077) != 0) {
    syslog(LOG_WARNING, "... insecure permissions %04o ...");
    if (st.st_uid != 0) { close(fd); return -1; }   /* only reached when perms are loose */
    /* root-owned + loose: warn but allow */
}
```

So a key file that was mode `0600` but owned by *any other uid* went through with **no check whatsoever** — an unprivileged user could supply the key material for the offline auth cache. Ownership is now checked unconditionally, plus `S_ISREG`.

### Should a root-owned 0644 key file be rejected outright?

**Yes — fail-closed**, and I checked the permissive path first:

- No test references a key file at all (`tests/test_auth_cache.c` and `tests/test_token_cache.c` only include `cache_key.h`); nothing depends on the warn-and-allow behaviour.
- The supported creation path, `scripts/ob-desktop-setup:400-403`, already does `chmod 600`.
- The only thing that produced a `0644` key file was the manual recipe in `SECURITY.md:743` — `dd if=/dev/urandom of=/etc/open-bastion/cache.key bs=32 count=1`, which under root's default umask 022 yields `0644 root`. That doc line is fixed in this PR to `chown root:root` + `chmod 600`, with the requirement spelled out.

The decisive point is that rejecting is **not** fail-closed in the dangerous sense: `cache_derive_key_with_keyfile()` (`src/cache_key.c:383-415`) treats an unreadable key file as absent, falls back to machine-id derivation and logs `cache key derived without key file (weaker security)`. The blast radius of a rejected key file is a weaker derivation and an invalidated offline cache (entries become undecryptable → cache miss → normal online auth), never a lockout. Given a group- or world-readable key file offers no confidentiality anyway, keeping it as "warn but use" bought nothing and made the syslog warning permanently ignorable.

## 3. `secret_store_get` has no output-buffer bounds check (#187)

`src/secret_store.c` read `ciphertext_len` from `st.st_size` and handed the caller's buffer straight to `EVP_DecryptUpdate()`; `secret_size` was used only for the trailing `explicit_bzero`. AES-256-GCM is a stream cipher — `EVP_DecryptUpdate` writes `ciphertext_len` bytes immediately, and the tag is only verified later in `EVP_DecryptFinal_ex()`. So an oversized secret file overflowed the caller's buffer **with unauthenticated attacker-controlled data**, before any integrity check could reject it.

Now rejected before the cipher is touched:

```c
if (ciphertext_len + TAG_SIZE > secret_size) { ... return -1; }
```

The `+ TAG_SIZE` margin is over-strict by 16 bytes for GCM (which needs no padding), but it costs nothing: there are no production callers of `secret_store_get()` in tree, and every test buffer is generously sized. It matches the `explicit_bzero(secret, secret_size)` already on the tag-mismatch path.

## 4. NSS file cache is world-readable (#189)

`nss/libnss_openbastion.c` created the cache with `mkdir(CACHE_DIR, 0755)` and `fchmod(..., 0644)`, so any local account could `ls /var/cache/nss_llng` and harvest name/uid/gid/gecos/home/shell for every SSO user the host had ever resolved. The code already acknowledges the sensitivity where it refuses to persist service accounts.

### Who reads it — and why 0600 is not possible

I traced the readers before touching the mode, and **0600 root would break the product**:

- An NSS module is loaded into **the calling process**, not a daemon. `_nss_openbastion_getpwuid_r` runs with the privileges of whoever called `getpwuid()`.
- `load_config()` returns failure unless it can read the root-only server token (`nss/libnss_openbastion.c:577`), and `query_llng_userinfo()` bails immediately without one. An unprivileged process therefore **can never query LLNG** — this file cache is its *only* source of uid → name data (`_nss_openbastion_getpwuid_r` has no other fallback: `file_cache_load_by_uid` fails → `NSS_STATUS_NOTFOUND`).
- OpenSSH's client calls `getpwuid(getuid())` at startup and **refuses to run** if it fails (`You don't exist, go away!`). A 0600 cache would break `ob-ssh` / `ob-scp` / `ob-sftp` for every bastion user, plus `ls -l`, `ps`, `id`, `w`.

I also considered and rejected `0600` owned by the cached user: the owner could then rewrite their own `gid`/`home`/`shell` in the entry (the loader validates none of it) and feed attacker-chosen passwd data back to **root** on the next lookup — a privilege escalation, strictly worse than the leak.

### What was done instead

- **Directory `0711` root** — traversable but not listable. Bulk enumeration of the SSO directory is gone; this is the same trick `ob-bastion-setup:722` already applies to `/etc/open-bastion` ("traversable (not readable) by users"). The `chmod` is explicit and unconditional because `mkdir()`'s mode is umask-masked and NSS runs under arbitrary umasks — and so an upgrade also tightens a `0755` directory left by an earlier version.
- **Entries stay `0644` root**, for the reasons above. The content is passwd-shaped — exactly what `/etc/passwd` exposes world-readably on every Unix — and the genuinely sensitive service accounts are already never written here.
- **`O_NOFOLLOW` + `O_CREAT|O_TRUNC` via `open()`** instead of `fopen("w")`: this is root truncating a path built from server-supplied data, so it must never be usable as a symlink write primitive.
- **`geteuid() == 0` guard**: only root may populate a cache the whole host trusts.

Net effect: the fix that was available is the enumeration surface, not confidentiality of an individual entry — the latter is structurally incompatible with serving `getpwuid()` to unprivileged callers. Flagging this explicitly since it is narrower than the issue asked for.

## 5. NSS module small defects (#196)

- **GID not range-checked.** The uid was validated against `[min_uid, max_uid]` and `NOBODY_UID`; the gid was taken verbatim. An unvalidated server-supplied gid is the more dangerous of the two — gid 0, or any system group such as `sudo`, `adm`, `docker`, `shadow`, would silently grant that group's privileges on every host. Now validated against the same range. **Deviation:** the uid path *fails the lookup* on an out-of-range value; for the gid I fall back to the locally configured `default_gid` with a warning instead, reusing the fallback the function already has for a missing or non-integer gid. Failing the whole lookup would turn a bad portal answer into a host-wide lockout, and there is no separate `min_gid`/`max_gid` setting to validate against.
- **Authorization header silently truncated.** `char auth_header[512]` vs. a token read into an 8192-byte buffer: any token over ~490 chars (routine for a JWT) produced a truncated `Bearer` value, a permanent 401, and a token reload burned on every lookup — with nothing in the logs saying why. The buffer is now sized for the largest token `load_server_token()` will accept, and the `snprintf()` return is checked so a future overrun fails loudly instead of 401-ing. Also fixes the `req_json` leak on that new error path and zeroes the token-bearing stack buffer after use.
- **`trim()` UB on the empty string.** It formed `str + strlen(str) - 1` unconditionally — a pointer before the start of the object — for an empty or all-blank value, which is what a `key =` line in `nss_openbastion.conf` yields. Guarded, matching the shape of `str_trim()` in `src/str_utils.c:21`.

## Verification

```
cmake -S . -B build-verify -DBUILD_TESTING=ON
cmake --build build-verify -j$(nproc)
ctest --test-dir build-verify --output-on-failure
```

Baseline before the change: clean build, 19/19. After:

```
      Start  1: test_config
      Start  2: test_rate_limiter
 2/19 Test  #2: test_rate_limiter ................   Passed    0.00 sec
      Start  3: test_audit_log
 3/19 Test  #3: test_audit_log ...................   Passed    0.05 sec
      Start  4: test_client_context
 4/19 Test  #4: test_client_context ..............   Passed    0.01 sec
      Start  5: test_secret_store
 5/19 Test  #5: test_secret_store ................   Passed    0.33 sec
      Start  6: test_nss_openbastion
 6/19 Test  #6: test_nss_openbastion .............   Passed    0.00 sec
      Start  7: test_nss_cache
 7/19 Test  #7: test_nss_cache ...................   Passed    0.00 sec
      Start  8: test_path_validator
 8/19 Test  #8: test_path_validator ..............   Passed    0.00 sec
      Start  9: test_username_validator
 9/19 Test  #9: test_username_validator ..........   Passed    0.00 sec
      Start 10: test_ob_cert_proto
10/19 Test #10: test_ob_cert_proto ...............   Passed    0.00 sec
      Start 11: test_ob_client
11/19 Test #11: test_ob_client ...................   Passed    0.00 sec
      Start 12: test_auth_cache
12/19 Test #12: test_auth_cache ..................   Passed    2.29 sec
      Start 13: test_token_manager
13/19 Test #13: test_token_manager ...............   Passed    0.01 sec
      Start 14: test_crowdsec
14/19 Test #14: test_crowdsec ....................   Passed    0.00 sec
      Start 15: test_service_account
15/19 Test #15: test_service_account .............   Passed    0.00 sec
      Start 16: test_token_cache
16/19 Test #16: test_token_cache .................   Passed    2.18 sec
      Start 17: test_token_manager_io
17/19 Test #17: test_token_manager_io ............   Passed    0.00 sec
      Start 18: test_notify
18/19 Test #18: test_notify ......................   Passed    4.01 sec
      Start 19: test_ssh_key_policy

Total Test time (real) =   8.91 sec
```

### New regression tests

`tests/test_nss_openbastion.c` works on *replicated copies* of the NSS helpers, so it cannot cover changes to the module itself. `tests/test_nss_cache.c` does `#include "../nss/libnss_openbastion.c"` and drives the real statics — that is where the NSS tests went. `CACHE_DIR` is now `#ifndef`-guarded so the test can redirect the file cache to `/tmp` instead of touching `/var/cache/nss_llng`.

Three cases added, **each confirmed to fail against the unfixed code**:

| Test | Covers | Unfixed result |
|---|---|---|
| `test_secret_store: output_buffer_too_small` | #187 — 512-byte secret into a 64-byte buffer, canary bytes checked | `FAIL` (11/12) — canaries clobbered |
| `test_nss_cache: file_cache_save_privileges` | #189 — non-root `file_cache_save()` must create nothing; modes asserted when run as root | `FAIL` — `unprivileged file_cache_save created ...` |
| `test_nss_cache: trim_empty_is_not_ub` | #196 — `trim("")`, `trim("   \t")` | passes trivially; documents intent, catches regressions under UBSan |

## Deployment note

`ob-cert-daemon` now refuses `/etc/open-bastion/openbastion.conf` and the server token unless they are `0600 root`, and `read_key_file` now ignores a `cache.key` that is not `0600 root`. Every installer already produces those modes; a host hand-configured otherwise will see an explicit journal/syslog line rather than a silent downgrade.

Fixes #184
Fixes #185
Fixes #187
Fixes #189
Fixes #196


nss/libnss_openbastion.c
```diff
@@ -1255,10 +1344,26 @@ static int query_llng_userinfo(const char *username, struct passwd *pw,
...
+     * the more dangerous of the two: gid 0 (or any system group such as sudo,
+     * adm, docker, shadow) would silently grant the SSO user that group's
+     * privileges on every host. Out-of-range values fall back to the locally
+     * configured default_gid rather than failing the lookup, so a bad answer
+     * from the portal degrades instead of locking users out. */
     if (json_object_object_get_ex(json, "gid", &val) &&
         json_object_is_type(val, json_type_int)) {
-        pw->pw_gid = (gid_t)json_object_get_int(val);
+        gid_t gid = (gid_t)json_object_get_int(val);
+        if (gid < (gid_t)g_config.min_uid || gid > (gid_t)g_config.max_uid ||
```
guimard: [Revue — MAJEUR] La validation du GID réutilise la plage UID synthétique [10000, 60000] : tout `gidNumber` LDAP classique (100, 1000, 5000…) est hors plage et remplacé silencieusement par `default_gid=100` quand l'admin exporte `gid` (chemin vivant via `pamAccessExportedVars`). Avant la PR : `pw_gid = gidNumber` (ex. 1000, groupe privé Debian) ; après : groupe primaire 100 → changement silencieux des permissions sur fichiers partagés, non documenté. Incohérent avec l'UID qui, lui, fait échouer tout le lookup. Suggestion : borner spécifiquement les gid (rejeter gid 0 et groupes système, ex. < 100, ou `min_gid`/`max_gid` configurables) — bloquer gid 0/sudo/docker est justifié, mais pas tous les groupes utilisateurs.

src/secret_store.c
```diff
@@ -469,6 +469,27 @@ int secret_store_get(secret_store_t *store,
     unsigned char *ciphertext = data + IV_SIZE;
     unsigned char *tag = data + IV_SIZE + ciphertext_len;
 
+    /*
```
guimard, line 485: [Revue — Mineur] `ciphertext_len + TAG_SIZE > secret_size` est sur-conservateur : en AES-256-GCM `EVP_DecryptFinal_ex` n'écrit jamais de plaintext supplémentaire (pas de padding). Un appelant qui alloue exactement `plaintext_len` (contrat naturel avec `actual_len`) est refusé. Le message d'erreur imprime `ciphertext_len > secret_size` alors que la condition teste `+ TAG_SIZE`. Impact nul aujourd'hui (seul tests/test_secret_store.c appelle), mais le défaut de contrat est verrouillé dans l'API. Suggestion : `ciphertext_len > secret_size` + message cohérent.

src/cache_key.c
```diff
@@ -331,17 +331,41 @@ static int read_key_file(const char *path, unsigned char *buf, size_t buf_size)
         return -1;
     }
 
-    /* Enforce strict permissions on key file */
+    /* Must be a regular file, not a fifo/device that could block or feed us
```
guimard, line 334: [Revue — Mineur] Rejet fail-closed de `cache.key` avec bits group/other (volontaire et correct), MAIS l'ancienne reco SECURITY.md (`dd if=/dev/urandom of=/etc/open-bastion/cache.key …` sans chmod) a produit des fichiers 0644 chez les déploiements existants. Post-PR : clé rejetée → dérivation par machine-id → toutes les entrées offline existantes deviennent indéchiffrables (ré-auth en ligne une fois pour les desktop SSO). Dégradation assumée et commentée, mais aucune note CHANGELOG/release dans la PR. Suggestion : entrée CHANGELOG explicite + syslog directif (« re-créer la clé avec chmod 600 »).

guimard (commented):
Revue de code — verdict : **À AJUSTER AVANT FUSION** (1 majeur, détail inline).

Le durcissement est de bonne facture : `open_checked()` conforme aux conventions de config.c/NSS, `trim()` UB corrigé, `auth_header` borné avec échec bruyant, cache NSS 0711 justifié, fail-closed cache.key cohérent avec les installateurs. Mais la borne GID casse silencieusement les groupes primaires exportés.

Nits :
- tests/test_nss_cache.c:23 — chemin prévisible `/tmp/test_nss_openbastion_cache` (tourne en root en CI) ; suggérer `mkdtemp`.
- `read_key_file` ne vérifie pas `S_ISUID`/`S_ISGID` (sans conséquence pratique, trivial à ajouter).

guimard:
Merci pour la revue — les 5 points sont traités, 5 commits sur la branche (`24308ca..6d29dfe`). CI verte (13/13).

---

## MAJEUR — la borne GID (`nss/libnss_openbastion.c`)

Vous avez raison, et la remarque va plus loin que le symptôme : réutiliser `[min_uid, max_uid]` était une erreur de catégorie. Ces bornes n'existent que pour que `generate_unique_uid()` frappe des UID synthétiques qui ne peuvent pas entrer en collision avec les comptes locaux ; un `gidNumber` LDAP n'a aucune raison d'y vivre. Résultat : 100, 1000, 5000 → `default_gid`, silencieusement.

Les GID ont maintenant **leur propre plage de politique**, `min_gid` / `max_gid` dans `nss_openbastion.conf`, par défaut `[1000, 65533]` (parseur, `config/nss_openbastion.conf.example`, `doc/admin-guide.md`, `doc/bastion-architecture.md`, `SECURITY.md`, `doc/security/00-architecture.md`).

### Pourquoi 1000 et pas 100

Vous proposiez « < 100 » comme coupe défendable en invitant à vérifier les conventions. Vérification faite, **1000 est la bonne ligne** :

| Source | Frontière |
|---|---|
| Debian `login.defs` / `adduser` | `SYS_GID_MIN=100`, `SYS_GID_MAX=999`, `GID_MIN=1000` |
| RHEL/Fedora `login.defs` | groupes système ≤ 999, `GID_MIN=1000` |
| ce module lui-même | `generate_unique_uid()` refuse déjà `min_uid < 1000` « to protect system UIDs » |

Surtout, **« < 100 » raterait précisément l'exemple que vous citez** : `sudo`=27, `adm`=4, `shadow`=42, `wheel`=10 sont bien < 100, mais `docker` (comme `lxd`, `libvirt`) est alloué dynamiquement par `adduser --system` dans **100–999** — typiquement 998 sur Debian. Or `docker` est un équivalent-root : c'est exactement le groupe que le contrôle doit bloquer.

Et rien de légitime n'est perdu :
- `gidNumber ≥ 1000` (le cas normal, groupe privé Debian inclus) passe **verbatim** ;
- `gid 100` (`users`) est déjà la valeur de `default_gid` par défaut : un site qui l'exporte obtient **la même valeur qu'avant**, aucun changement observable ;
- un site qui exporte réellement un gidNumber bas règle `min_gid = 100` ;
- `gid 0` et `nogroup` (65534) sont refusés **quoi que dise la configuration**, et une plage incohérente (`max_gid = 0`, `min_gid > max_gid`) retombe sur la politique compilée plutôt que sur « tout refuser ».

### Repli plutôt qu'échec — et assumé, pas silencieux

J'ai gardé le repli sur `default_gid`, contrairement à l'UID qui fait échouer le lookup. L'asymétrie est délibérée :

- **L'UID est une identité.** Un mauvais UID = mauvaise propriété de fichiers et mauvaise attribution d'audit. Refuser de répondre est la bonne réaction.
- **Le GID ne sélectionne qu'un jeu d'accès.** Faire échouer le lookup transformerait *une* entrée `pamAccessExportedVars` erronée en `NSS_STATUS_NOTFOUND` pour **tous** les utilisateurs SSO de l'hôte — et OpenSSH refuse carrément de démarrer quand `getpwuid()` échoue (« You don't exist, go away! »). Une panne de portail deviendrait un lockout de flotte.
- La fonction **avait déjà** ce repli pour un `gid` absent ou non entier ; un gid hors politique est le même cas (« le serveur n'a pas fourni de gid utilisable »).

Votre objection portait sur le **silence**, pas sur le repli. Le repli journalise désormais la valeur fautive, l'utilisateur, la plage active et le remède :

```
libnss_openbastion: server-supplied gid 27 for user alice is outside the allowed
group range [1000, 65533] (or is a reserved gid) - using default_gid 100 instead;
adjust min_gid/max_gid in nss_openbastion.conf if this gid is legitimate
```

### Test de non-régression

La sélection est extraite en `select_primary_gid()` / `gid_in_policy()` pour être atteignable depuis les tests ; `tests/test_nss_cache.c` pilote la **fonction livrée** via le vrai `g_config` (pas une copie). 15 cas : `gidNumber` 1000 / 5000 / 50000 / 65533 survivent tels quels ; gid 0, 27 (sudo), 10 (wheel), 998 (docker dynamique), 65534, négatif → `default_gid` ; absent / chaîne → repli historique ; `min_gid=0` ne veut pas dire « tout refuser » ; `min_gid=100` admet 100 mais refuse toujours 27.

Vérifié en le cassant volontairement — contre l'ancienne borne :

```
    gidNumber 1000 survives: got gid 100, expected 1000
    gidNumber 5000 survives: got gid 100, expected 5000
    max_gid boundary survives: got gid 100, expected 65533
    min_gid=100 admits gid 100: got gid 65533, expected 100
  Testing server_gid_policy... FAIL
```

---

## Mineur 1 — borne `secret_store_get` (`src/secret_store.c:472`)

Corrigé, l'analyse est juste : GCM est non paddé, `EVP_DecryptFinal_ex()` n'écrit aucun plaintext supplémentaire, donc `ciphertext_len` **est** la longueur totale du clair. La borne est maintenant `ciphertext_len > secret_size` et le message correspond à la condition testée (il imprimait bien `ciphertext_len > secret_size` alors que le test portait sur `+ TAG_SIZE`). Le garde-fou anti-débordement lui-même est inchangé.

Nouveau test `output_buffer_exact_size` : secret de 64 octets déchiffré dans exactement 64 octets utilisables, canaris autour + comparaison du contenu. Il **échoue** contre l'ancienne borne (12/13). C'est bien parce que l'impact est nul aujourd'hui qu'il fallait corriger maintenant, avant de verrouiller le mauvais contrat.

## Mineur 2 — note de release `cache.key` (`src/cache_key.c:334`)

Les deux ajouts demandés :

1. **`CHANGELOG.md`, section `[Unreleased]` / `### Changed`** : qui est concerné (les hôtes montés depuis l'ancienne recette `dd …` de SECURITY.md, 0644 sous umask 022 — `ob-desktop-setup` a toujours créé la clé en 0600, donc non concerné), l'effet exact (clé ignorée → dérivation par machine-id → entrées offline existantes indéchiffrables = **cache miss, pas lockout**, une ré-auth en ligne pour les utilisateurs desktop SSO, aucun nettoyage manuel), et le remède qui restaure la dérivation forte.
2. **Message syslog directif** au point de rejet, avec la commande exacte :

```
open-bastion: key file /etc/open-bastion/cache.key has insecure permissions 0644
(must be 0600 root:root) - rejected, falling back to machine-id key derivation;
existing offline cache entries become undecryptable. Fix with:
chown root:root /etc/open-bastion/cache.key && chmod 600 /etc/open-bastion/cache.key
```

## Nits

- **`mkdtemp`** : `CACHE_DIR` est désormais *valué par fonction* et résout un `mkdtemp()` par exécution, nettoyé en fin de `main()`. **Volontairement aligné sur `fix/nss-name-file-cache` (#174)** que j'ai lu avant : mêmes noms d'aides (`test_base` / `test_cache_root` / `rm_rf`), même gabarit `/tmp/ob_nss_cache_test_XXXXXX`, même disposition `<base>/cache`, et la racine du cache n'est volontairement pas pré-créée. Voir la note de conflit ci-dessous.
- **`S_ISUID`/`S_ISGID`** dans `read_key_file` : ajouté, refus explicite avec log. setuid sur un fichier de clé n'a aucun sens (c'est de la donnée) et signale un fichier déposé plutôt que généré par `ob-desktop-setup`.

---

## Conflit avec #174 (`fix/nss-name-file-cache`)

À signaler franchement : les deux branches **conflictaient déjà** avant cette revue — 9 régions de conflit entre `55a6d91` et `origin/fix/nss-name-file-cache`, toutes issues de la réécriture propre à #174 (`CACHE_DIR_BYNAME`, `CACHE_TRUSTED_UID`, la refonte de `file_cache_save`, la macro `RUN()`). Mes commits en ajoutent **une seule** : le bloc de test `server_gid_policy`, purement additif (résolution = garder les deux côtés). Le passage `mkdtemp` **rapproche** au contraire les deux fichiers : l'en-tête « Harness overrides » ne diffère plus que par la formulation du commentaire. Aucun conflit avec `main` (`git merge-tree` : 0), donc pas de rebase.

## Vérification

| Configuration | Résultat |
|---|---|
| `cmake -S . -B build-verify -DBUILD_TESTING=ON` | **19/19**, 0 warning |
| `-DINSTALL_DESKTOP=ON` | **21/21** |
| `-DCMAKE_BUILD_TYPE=Debug -DUSE_LIBSODIUM=ON -fsanitize=address,undefined` | **19/19** |
| `gh pr checks 222` | **13/13 pass** (dont CodeQL, Build Debug with Sanitizers, Integration Tests Docker + Mode E, RPM Rocky 9) |

`test_secret_store` passe de 12 à 13 cas, `test_nss_cache` de 4 à 5.

PR #223, opened 2026-09-05
fix(concurrency): lock lockout counters and stop sharing state temp files

### LINKED ISSUE #186: Lockout counters (rate limiter, offline cache) have no inter-process lock: parallel attempts bypass the threshold
Two lockout counters are read-modify-write across processes with no lock, so parallel attempts lose increments and can outrun the threshold.

## 1. Rate limiter

`src/rate_limiter.c:340-402` (`rate_limiter_record_failure`):

```c
if (!load_state(path, &state)) { memset(&state, 0, sizeof(state)); ... }
state.failure_count++;
...
save_state(path, &state);     /* no flock/lockf anywhere in the file */
```

`grep -n "flock\|lockf\|fcntl" src/rate_limiter.c` returns nothing. `save_state` uses `O_EXCL` + `renameat` so it is symlink-safe, but that gives no logical exclusivity: N concurrent PAM processes all read the same `failure_count` and each writes `k+1`, so the counter advances by 1 instead of N.

Practical amplification is bounded by sshd's `MaxStartups` and, in Mode E, by the fact that the only consumer of `pam_sm_authenticate` is `sudo` (SSH takes the `authorize_only` path), which requires an existing local shell.

## 2. Offline cache lockout

`src/offline_cache.c:1179-1204` has the identical shape on `failed_attempts`, re-reading, decrypting, mutating and re-encrypting the whole entry with no inter-process lock. The 5-attempt lockout can be blown past by parallelising guesses.

Throughput there is bounded by Argon2id at 64 MiB x 3 iterations per attempt, and the path is desktop/offline only, so local access is already required.

A repo-wide `grep -rn "flock\|lockf\|F_SETLK" src/ nss/` finds locking only at `src/pam_openbastion.c:3784,3789` (the `/etc/passwd` and `/etc/shadow` writes).

## Suggested fix

Take an `flock` around the read-modify-write in both places, and add concurrency tests — the existing suites are single-process throughout.

---

### LINKED ISSUE #197: Cache and secret-store temp files use a fixed name without O_EXCL/PID, unlike rate_limiter and cache_key
Two writers use a fixed `.tmp` filename with `O_TRUNC` instead of the PID + `O_EXCL` pattern already used elsewhere in the codebase.

## Affected

`src/token_cache.c:649-652` and `src/auth_cache.c:616-619` (identical):

```c
snprintf(temp_path, sizeof(temp_path), "%s.tmp", path);
int fd = open(temp_path, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0600);
```

`src/secret_store.c:359`: `snprintf(temp_path, sizeof(temp_path), "%s.tmp", path);`

## The correct pattern already exists

`src/rate_limiter.c:207-212` uses `"%s.tmp.%d", filename, getpid()` with `O_EXCL`, and `src/cache_key.c:209-216` does the same.

## Impact

Two concurrent writers interleave into one `O_TRUNC` file and publish a corrupt entry. For the caches, lookup detects this and deletes the entry, so it is fail-closed cache loss. For `secret_store_put`, two concurrent puts on the same key corrupt the stored secret. Not reachable by a non-root user — the directories are 0700 root.

## Suggested fix

Adopt the `rate_limiter.c` idiom (PID suffix + `O_EXCL`) in all three places.

---

### LINKED ISSUE #198: ob-record-sink: duplicate session_id truncates live metadata; no cumulative disk quota
Two issues in `src/ob-record-sink.c`.

## 1. A duplicate session_id truncates a live session's metadata

`src/ob-record-sink.c:349-355`: `write_metadata()` runs first and opens with `O_TRUNC` (`:225`), while the recording file is correctly opened `O_EXCL` (`:355`):

```c
int fd = openat(dirfd, name, O_WRONLY|O_CREAT|O_TRUNC|O_NOFOLLOW, 0640);   /* write_metadata */
...
write_metadata(dirfd, meta_name, session_id, ...);                          /* :349, runs first */
int recfd = openat(dirfd, rec_name, O_WRONLY|O_CREAT|O_EXCL|O_NOFOLLOW, 0640);  /* :355, rejects the dup */
```

The client-supplied `session_id` is validated for charset (`:312`) but not for uniqueness. A duplicate plus a same-second timestamp yields the same metadata filename, and the `O_TRUNC` clobbers the existing JSON *before* the `O_EXCL` rejects the duplicate connection. Since `dirfd` is the per-user session directory, the victim is always the same user — self-inflicted. One flag fixes it: `O_EXCL` on the initial metadata write.

## 2. No cumulative disk quota

`src/ob-record-sink.c:61` caps a recording at 1 GiB, and `:385-401` applies it per connection. There is no per-user or store-wide quota and no session-count limit, so N sequential 1 GiB streams are unbounded. `systemd/ob-session-prune.timer` bounds long-term growth (compression plus a 365-day default retention) but not a burst.

Note that the socket being world-connectable, and recordings being forgeable by their own purported owner, are **accepted** limits documented at `doc/design/tamper-evident-session-recording.md:312-320` — that part is not in scope here.

## Suggested fix

Add `O_EXCL` to the metadata write; add a per-user cumulative quota or document the disk-exhaustion bound explicitly.

---

guimard:
Three state files were written as if only one process could ever touch them, while PAM in fact runs one process per authentication attempt.

Fixes #186
Fixes #197
Fixes #198

## 1. Lockout counters had no inter-process lock (#186)

**Defect.** `rate_limiter_record_failure()` (`src/rate_limiter.c`) and `offline_cache_verify()` (`src/offline_cache.c`) both did an unguarded read-modify-write:

```c
if (!load_state(path, &state)) { memset(&state, 0, sizeof(state)); ... }
state.failure_count++;
save_state(path, &state);
```

`grep -n "flock\|lockf\|fcntl" src/rate_limiter.c` returned nothing. N concurrent PAM processes each read the same count `k` and each wrote `k+1`, so the counter advanced by 1 instead of N and `max_attempts` / `max_failed_attempts` could be outrun by simply attacking in parallel. `offline_cache.c` had the identical shape on `failed_attempts` (re-read, decrypt, mutate, re-encrypt the whole entry).

**Fix.** Both serialize the whole sequence with `flock(LOCK_EX)`, matching the idiom already used for the `/etc/passwd` / `/etc/shadow` writes in `src/pam_openbastion.c:3784,3789`.

- The lock lives in a **companion file**, not the state file itself: `save_state()` / `update_cache_entry()` publish by atomic `rename`, so the state inode changes on every update and a lock held on it would stop excluding a process that opened the new inode. The lock file is therefore never unlinked, and its name is chosen so the existing `".state"` / `".cred"` directory scans (cleanup, store quota, invalidate_all, stats) ignore it.
- The read-modify-write bodies were split into `*_locked()` helpers wrapped by the public functions, so **every** exit path — error paths included — releases the lock, with no new lock-ordering (a single lock, never nested).
- `rate_limiter_reset()` and `offline_cache_reset_failures()` take the same lock so a reset cannot land inside a concurrent increment.
- Failing to acquire the lock falls back to the previous unlocked behaviour rather than refusing the operation: an unwritable/symlinked lock path must not become a way to disable rate limiting or to deny offline logins. A symlink planted at the lock path is removed and retried once, as `load_state_at()` / `read_cache_entry()` already do.

## 2. Cache/secret temp files used a fixed name without O_EXCL (#197)

**Defect.** `src/token_cache.c`, `src/auth_cache.c` and `src/secret_store.c` all did:

```c
snprintf(temp_path, sizeof(temp_path), "%s.tmp", path);
int fd = open(temp_path, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0600);
```

Two concurrent writers interleave into one `O_TRUNC` file and rename the mixture into place — an unauthenticated blob that then fails AES-GCM on read.

**Fix.** All three adopt the correct pattern already present at `src/rate_limiter.c:207-212` and `src/cache_key.c:209-216`: `"%s.tmp.%d", path, getpid()` plus `O_EXCL`, with the stale-temp unlink-and-retry on `EEXIST` and a truncation check on the `snprintf`. `O_NOFOLLOW` and the atomic rename are unchanged; the temp buffers were widened for the pid suffix.

## 3. ob-record-sink metadata write was not O_EXCL (#198)

**Defect.** `write_metadata()` (`src/ob-record-sink.c`) opened with `O_TRUNC` while the recording file correctly used `O_EXCL` — and `write_metadata` runs **first**. A replayed client-supplied `session_id` plus a same-second timestamp truncated an existing session's metadata JSON before the `O_EXCL` on the typescript rejected the duplicate connection.

**Fix.** `write_metadata()` takes an `exclusive` flag, because it has two callers with opposite needs:

- the **initial** creation uses `O_EXCL` and returns `-2` on `EEXIST`, so `main()` refuses the session exactly like a duplicate recording (`goto reject_dir`);
- the **final** write (end time + observed `completed`/`truncated`/`aborted` status) still updates our own file in place, so the update path is untouched.

Other metadata write failures stay non-fatal, as before.

## Tests

Fork-based concurrency regressions — children are released together by a pipe barrier and the expected values are exact, so they are deterministic assertions rather than timing measurements:

- `tests/test_rate_limiter.c` — `concurrent_failures`: 8 processes x 25 failures must leave exactly 200.
- `tests/test_offline_cache.c` — `concurrent_failed_attempts`: 6 parallel wrong passwords must leave exactly 6.
- `tests/test_secret_store.c` — `concurrent_put`: concurrent writers must never publish an entry that fails to decrypt, and no shared temp file may be left behind.

Deterministic temp-name regressions (`temp_file_is_private` in `tests/test_token_cache.c`, `tests/test_auth_cache.c`, `tests/test_secret_store.c`): an obstacle planted at the old fixed `"<path>.tmp"` must be left strictly untouched by a store.

`tests/test_ob_record_sink.sh` — new case 5: a duplicate `session_id` in the same second must be refused with the pre-existing metadata intact.

**Every new test was also compiled against the pre-fix sources and fails there**, e.g.:

```
  Testing concurrent_failures... (count=31 expected=200) FAIL
  Testing concurrent_failed_attempts... (failed_attempts=1 expected=6) FAILED
  Testing temp_file_is_private... FAIL
FAIL duplicate session_id truncated the existing metadata (.../20260905-145813_dupid001.json)
```

## Verification

```
$ cmake --build build-verify -j$(nproc)      # clean, -Wall -Wextra -Werror, no warnings
Test project /home/yadd/dev/github/open-bastion/.claude/worktrees/agent-a39ba24cf967bfdaa/build-verify
      Start  1: test_config
      Start  2: test_rate_limiter
 2/19 Test  #2: test_rate_limiter ................   Passed    0.01 sec
      Start  3: test_audit_log
 3/19 Test  #3: test_audit_log ...................   Passed    0.05 sec
      Start  4: test_client_context
 4/19 Test  #4: test_client_context ..............   Passed    0.00 sec
      Start  5: test_secret_store
 5/19 Test  #5: test_secret_store ................   Passed    0.37 sec
      Start  6: test_nss_openbastion
 6/19 Test  #6: test_nss_openbastion .............   Passed    0.00 sec
      Start  7: test_nss_cache
 7/19 Test  #7: test_nss_cache ...................   Passed    0.00 sec
      Start  8: test_path_validator
 8/19 Test  #8: test_path_validator ..............   Passed    0.00 sec
      Start  9: test_username_validator
 9/19 Test  #9: test_username_validator ..........   Passed    0.00 sec
      Start 10: test_ob_cert_proto
10/19 Test #10: test_ob_cert_proto ...............   Passed    0.00 sec
      Start 11: test_ob_client
11/19 Test #11: test_ob_client ...................   Passed    0.00 sec
      Start 12: test_auth_cache
12/19 Test #12: test_auth_cache ..................   Passed    2.34 sec
      Start 13: test_token_manager
13/19 Test #13: test_token_manager ...............   Passed    0.00 sec
      Start 14: test_crowdsec
14/19 Test #14: test_crowdsec ....................   Passed    0.00 sec
      Start 15: test_service_account
15/19 Test #15: test_service_account .............   Passed    0.00 sec
      Start 16: test_token_cache
16/19 Test #16: test_token_cache .................   Passed    2.18 sec
      Start 17: test_token_manager_io
17/19 Test #17: test_token_manager_io ............   Passed    0.00 sec
      Start 18: test_notify
18/19 Test #18: test_notify ......................   Passed    4.01 sec
      Start 19: test_ssh_key_policy

Total Test time (real) =   9.00 sec
```

The offline-cache tests only build with `-DINSTALL_DESKTOP=ON` (that is why the default suite is 19 tests). With it:

```
$ cmake -S . -B build-desktop -DBUILD_TESTING=ON -DINSTALL_DESKTOP=ON
$ ctest --test-dir build-desktop --output-on-failure
100% tests passed, 0 tests failed out of 21
```

And the record-sink end-to-end script (not registered in ctest, like the other `.sh` tests):

```
$ bash tests/test_ob_record_sink.sh
ok   PTY stream captured (script -> FIFO -> sink)
ok   metadata json: user from SO_PEERCRED, status completed
ok   path derives from SO_PEERCRED uid (no header-controlled user)
ok   transfer session: metadata-only, empty placeholder
ok   fail-closed: unreachable sink → connector non-zero
ok   duplicate session_id refused, existing metadata intact
=== record-sink e2e: PASS ===
```


src/auth_cache.c
```diff
@@ -612,16 +612,34 @@ int auth_cache_store(auth_cache_t *cache,
     char path[PATH_MAX];
     build_auth_cache_path(cache, user, server_group, host, path, sizeof(path));
 
-    char temp_path[PATH_MAX + 8];
-    snprintf(temp_path, sizeof(temp_path), "%s.tmp", path);
-
-    int fd = open(temp_path, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0600);
-    if (fd < 0) {
+    /*
```
guimard, line 615: [Revue — Mineur] Le motif `EEXIST → unlink → retry` n'est sûr qu'en multi-processus : deux threads du même processus calculent le même `temp_path` (même PID) ; le second `unlink()` le temp du premier (encore ouvert en écriture) et le premier peut `rename()` le temp partiellement écrit du second. Conséquence bornée : fichier final tronqué → rejeté par HMAC/parse → cache miss transitoire, jamais de fausses données servies. Risque faible avec le modèle « un processus PAM par tentative », mais identique dans token_cache.c ~647-670 et secret_store.c ~358-393. Suggestion : suffixe aléatoire par tentative (ou O_TMPFILE + linkat), ou traiter EEXIST en échec propre sans unlink.

src/offline_cache.c
```diff
@@ -370,6 +371,68 @@ static void build_cache_path(offline_cache_t *cache, const char *user,
     snprintf(path, path_size, "%s/%s.cred", cache->cache_dir, hash);
 }
 
+/*
```
guimard, line 374: [Revue — Mineur] (a) `flock(LOCK_EX)` bloquant sans timeout : un détenteur gelé (SIGSTOP/freezer) fait pendre le processus PAM indéfiniment, ce qui contredit l'intention « degraded plutôt que refus » (répertoires 0700 root en config saine, donc non exploitable par un local user). Suggestion : LOCK_NB + retry borné puis fallback dégradé. (b) Les `.lock` ne sont jamais purgés (cleanup/stats filtrent `.cred`/`.state` mais pas `.lock`) → accumulation d'inodes vides par clé. Suggestion : purge opportuniste des `.lock` orphelins dans les fonctions de nettoyage.

guimard (commented):
Revue de code — verdict : **APPROUVÉ** (2 mineurs non bloquants, détail inline).

Bon choix de `flock` sur fichier compagnon (inode stable vs rename atomique), lock couvrant tout le read-modify-write, verrou relâché par le noyau sur crash, fallback dégradé documenté plutôt que refus de service, temps per-PID + O_EXCL + O_NOFOLLOW éliminent l'entrelacement multi-processus (#197), O_EXCL sur ob-record-sink empêche la troncature de métadonnées (#198). Tests de concurrence réels (fork, barrières pipe, assertions exactes).

Nit : course préexistante `strstr(".cache"/".authcache")` vs temps per-PID dans les balayages de nettoyage — pas une régression.

Issue #151, opened 2026-06-16
User can delete its own records


guimard:
User should not have the right to list `/var/lib/open-bastion/sessions/<username>`

PR #226, opened 2026-09-05
fix: low-severity hardening batch — C core, packaging/systemd, demo hygiene

### LINKED ISSUE #199: Assorted C hardening: CURLOPT_NOSIGNAL, audit_log fchmod/ident, authorize parsing, state-dir checks, token inode re-adoption
Small hardening and consistency items found during an audit pass, grouped because each is a few lines.

## 1. `CURLOPT_NOSIGNAL` is not set in `ob_client`

`src/ob_client.c:604-657` (`setup_curl`) sets `TIMEOUT`, `SSLVERSION`, `PINNEDPUBLICKEY`, `TCP_KEEPALIVE` and `ACCEPT_ENCODING`, but not `CURLOPT_NOSIGNAL`. The NSS module does set it (`nss/libnss_openbastion.c:1147`). Without it, libcurl may use `SIGALRM`/`siglongjmp` for name-resolution timeouts, which is unsafe in a multi-threaded host process such as the LightDM greeter.

## 2. `audit_log` re-opens a stricter log file to 0640

`src/audit_log.c:285-308` enforces `O_NOFOLLOW`, ownership and no group/other write — good — but then calls `fchmod(fd, 0640)` unconditionally on every event, which re-opens an administrator's deliberate 0600 log to the group. There is also no size cap or rotation.

## 3. Audit events are logged under a stale syslog ident

`src/audit_log.c:111` calls `openlog("pam_llng", ...)`, the pre-0.2.1 name, so the `grep pam_openbastion` recipes in `doc/troubleshooting.md` miss audit events.

## 4. `/pam/authorize` parsing is stricter than `/pam/verify`

`src/ob_client.c:1169-1175` rejects a response without a `user` field as a server error, whereas `/pam/verify` was fixed in #177 to surface a negative verdict cleanly (`:803-813`). The LLNG plugin does always send `user` on a negative authorize verdict (`PamAccess.pm:415-418`, `:465-468`, `:523-526`), so no deployed server triggers this — but the asymmetry is worth removing. The only real loss would be audit accuracy (`AUDIT_SERVER_ERROR` instead of `AUDIT_AUTHZ_DENIED`); it stays fail-closed either way.

## 5. Pre-existing state directories are not permission-checked

`src/auth_cache.c:295-303`, `src/token_cache.c:260-268` and `src/rate_limiter.c:277-292` all `mkdir(cache_dir, 0700)` but never verify the mode or owner of a directory that already exists. `src/cache_key.c:192-200` opens the salt file `O_NOFOLLOW` but does not check its mode or owner either. The packaging sets the parent to 0700 root (`debian/open-bastion.postinst:157-158`), which shields a lax subdirectory, so this is defence in depth.

## 6. On-demand token refresh does not re-adopt the new inode

`src/pam_openbastion.c:309-313` refuses to continue when the token file inode changes. `data->token_file_inode` is set once at `:1322-1324` and never updated — including by `refresh_server_token_on_demand()` (`:2179-2243`), which persists via `rename()`. Since `ob-heartbeat` rotates the token every 5 minutes (`systemd/ob-heartbeat.timer:12`), exactly `TOKEN_RECHECK_INTERVAL`, a long-lived PAM handle can lock itself out. No such consumer exists in the core build (sshd and sudo create a fresh handle per invocation), so this is latent rather than reachable — but the on-demand refresh self-poisoning the handle it just healed is worth fixing.

---

### LINKED ISSUE #200: Packaging/systemd: RPM socket re-arm parity, sudoers visudo validation, debconf db_reset, portal_url validation, unit hardening
Packaging and systemd items found during an audit pass.

## 1. RPM `%post` has no equivalent of the Debian socket re-arm

`rpm/open-bastion.spec:171-223` runs `%systemd_post` only for `ob-heartbeat.timer` and `ob-session-prune.timer`; the sockets appear only in `%preun`/`%postun`. `debian/open-bastion.postinst:284-298` has the role-gated `systemctl enable --now ob-cert.socket` / conditional `ob-record.socket` added in 0.6.1.

**Scope correction**: this is *not* an active RPM regression. Nothing in the RPM scriptlets disables the sockets on upgrade — `%systemd_preun` is a no-op when `$1 >= 1`, and `%systemd_postun_with_restart` only `try-restart`s. The Debian breakage came from the `deb-systemd-helper` state layer, which has no RPM analogue. The value of porting the block is repair-on-upgrade and migration from pre-socket versions.

## 2. `ob-bastion-setup` writes a sudoers file without validating it

`scripts/ob-bastion-setup:952-961` writes `/etc/sudoers.d/open-bastion` with bare `echo` and `chmod`, no `visudo -cf`. `scripts/ob-backend-setup:607-631` does it correctly, with `visudo -cf` then `install -m 0440`. A malformed sudoers drop-in breaks `sudo` host-wide.

## 3. The debconf client secret is cleared conditionally and never reset

`debian/open-bastion.postinst:251-253` calls `db_set open-bastion/client-secret ""` inside the `if [ "$MODE" != "none" ]` block opened at `:180`. Choosing `pam-mode=none` after answering the secret question therefore leaves the value in place, and `db_set ""` keeps the question flagged as seen where `db_reset` would restore the default. Impact is bounded — `/var/cache/debconf/passwords.dat` is root-only and the template is correctly `Type: password` (`debian/open-bastion.templates:24-27`) — but an unconditional `db_reset` at the end of `configure` is a one-line fix.

## 4. `portal_url` is embedded unvalidated into generated root scripts

`scripts/ob-bastion-setup:877-891` and `scripts/ob-backend-setup:717-729` write `${PORTAL_URL}` into `/usr/local/bin/open-bastion-refresh-krl` through an unquoted heredoc, and that script runs from root cron every 30 minutes. `PORTAL_URL` gets only an emptiness check and a trailing-slash strip (`ob-bastion-setup:1952-1968`).

No privilege boundary is crossed — the setup already requires root and the value is the admin's own `--portal` argument — and via `ob-builder` the value is already sanitised by `is_valid_url`. This is a parity gap, not the root RCE it might look like. Worth applying the same regex for consistency.

## 5. Residual systemd hardening

`systemd/ob-session-monitor.service` has no `ProtectSystem=`; it legitimately writes `/etc/open-bastion/force_online`, so `ProtectSystem=full` plus the existing `ReadWritePaths` would work. `systemd/ob-cert@.service` and `ob-record@.service` lack `UMask=0077` and `SystemCallFilter=@system-service`; both run `User=root` by design, so the rest is advisory.

---

### LINKED ISSUE #201: Demo/lab hygiene: cookie file pattern, world-writable recordings tree, obsolete lab-curl-shim.sh
Hygiene items in the demo and lab trees. None reaches a production code path, but these entrypoints are the templates people copy.

## 1. Fixed-path cookie file, dumped to the log on failure

`docker-demo-cert/bastion/entrypoint.sh:168-169,186-191` uses `COOKIE_FILE="/tmp/admin_cookies"` with default permissions, leaves it on disk, and `cat`s it to the container log on failure. The same fixed path appears in all 8 demo entrypoints (cert, maxsec, token, token-svc x bastion/backend).

`quick-start/server/entrypoint.sh:39-41` already has the correct pattern: `mktemp` + `chmod 600` + `trap`.

## 2. World-writable recordings tree

`docker-demo-cert/bastion/entrypoint.sh:377-378` and `docker-demo-token/bastion/entrypoint.sh:282-283` do `chmod 1777 /var/lib/open-bastion/sessions`. The maxsec and token-svc demos use `1770`.

Production has moved on twice: the current layout is `root:ob-sessions 0750` (`debian/open-bastion.postinst:172-173`, the RPM `%post`, and `src/ob-record-sink.c:20`), so even the `1770` demos are stale, and `1777` directly contradicts the tamper-evident design.

## 3. Obsolete lab curl shim still in the tree

`local/ansible/roles/open-bastion/files/lab-curl-shim.sh` is a curl shim that spoofs `/ssh/ca`. `tasks/main.yml` now removes the shim at deploy time, so the file is dead code with a nasty shape.

## Not in scope

The other lab defaults (`StrictHostKeyChecking=no` in the demos, `deb [trusted=yes]` in `local-test/lib.sh:172`, `verify_ssl=false` in `local/ansible/gen-backend/inv-lab.yml`) are all confined to demo/lab trees and each carries an explicit in-file comment or README note saying so. The static demo LLNG server token in `docker-demo-*/sso/init-tokens.sh` is a fixture for a throwaway container alongside public `dwho/dwho` credentials.

## Suggested fix

Adopt the quick-start cookie pattern in the demo entrypoints, align the recordings tree with the production 0750 `root:ob-sessions` layout, delete `lab-curl-shim.sh`.

---

guimard:
A batch of low-severity defects across the C core, packaging/systemd, and the demo containers. Three commits, one per issue, reviewable independently.

Fixes #199
Fixes #200
Fixes #201

## A — C hardening (#199), commit 1

| | |
|---|---|
| **A1** | `setup_curl()` never set `CURLOPT_NOSIGNAL`, so libcurl may arm SIGALRM and `siglongjmp` out of the DNS resolver — undefined behaviour inside the multi-threaded processes we are loaded into, notably the LightDM greeter. The NSS module already set it; the two are now aligned. |
| **A2** | `fchmod(fd, 0640)` ran on **every** audit event, so a log an administrator deliberately tightened to 0600 was re-opened to the group on the next login. The file is now opened `O_EXCL` first to detect creation and the mode is forced only when we created it (which still defeats the umask). All existing `O_NOFOLLOW` / ownership / group-write checks untouched. |
| **A2 (rotation)** | Chose **logrotate template + soft size warning** over in-module rotation: every `sshd` and `sudo` process on the host appends to the same file, so a module that renamed or truncated it would race with its own peers. Ships `config/logrotate/open-bastion` to `/usr/share/open-bastion/logrotate/` — the same read-only, non-conffile arrangement as the existing auditd templates, because `audit_log_file` is configurable — plus a one-shot syslog warning past 100 MB for hosts that never install it. No audit event is ever dropped. |
| **A3** | `openlog()` still said `pam_llng` (pre-0.2.1), so audit events did not appear under `journalctl -t pam_openbastion` alongside the module's own `pam_syslog` messages. The JSON `module` field said the same. Both fixed; `doc/troubleshooting.md` and `doc/security.md` gained the matching grep recipes (there were none for audit events before). |
| **A4** | `/pam/authorize` reported a response without `user` as a server error → `AUDIT_SERVER_ERROR` / `PAM_AUTHINFO_UNAVAIL`, which also wrongly enters the offline-cache fallback, instead of the `AUDIT_AUTHZ_DENIED` / `PAM_PERM_DENIED` a negative verdict deserves — the same defect fixed for `/pam/verify` in #177. No deployed server triggers it (`PamAccess.pm` always sends `user` on a denial), so this only hardens the contract; fail-closed either way. |
| **A5** | **Skipped — remains open under #199.** Another agent is concurrently editing `auth_cache.c`, `token_cache.c` and `rate_limiter.c`, so permission-checking a pre-existing state directory was left out to avoid conflicting work. |
| **A6** | `verify_token_file_security()` treats an inode change as a security violation, but `token_file_inode` was recorded once at load while the token is legitimately replaced by `rename()` — by `ob-heartbeat` every 5 minutes, i.e. exactly `TOKEN_RECHECK_INTERVAL`, and by `refresh_server_token_on_demand()` itself. A long-lived PAM handle could lock itself out, with the on-demand refresh poisoning the handle it had just healed. Latent (no such consumer in the core build), fixed anyway. |

## B — packaging and systemd (#200), commit 2

| | |
|---|---|
| **B1** | Ported the Debian postinst socket re-arm to `%post`. Written as **repair/migration parity, not a regression fix**: nothing in the RPM scriptlets disables the sockets on upgrade (`%systemd_preun` is a no-op while `$1 >= 1`), and the Debian breakage came from the `deb-systemd-helper` state layer, which has no RPM analogue. What it buys is the same self-healing on both distros. Guarded on the bastion-role sshd drop-in, so it never touches a backend or an unconfigured host. |
| **B2** | `ob-bastion-setup` wrote `/etc/sudoers.d/open-bastion` with bare `echo` + `chmod`, publishing it before anything checked it — and a malformed drop-in breaks `sudo` host-wide. Now temp file → `visudo -cf` → `install -m 0440` with the result checked, matching `ob-backend-setup`. |
| **B3** | `db_set open-bastion/client-secret ""` sat inside the `if [ "$MODE" != "none" ]` block, so `pam-mode=none` left the secret in `passwords.dat` forever — precisely the case where it is never used. Now unconditional, and `db_reset` rather than `db_set ""` (which leaves the question flagged as seen with a blank answer, so a later reconfigure silently reuses the blank). No rebase was needed: `origin/main` had not moved. |
| **B4** | `PORTAL_URL` is interpolated into `/usr/local/bin/open-bastion-refresh-krl` through an unquoted heredoc, and that script runs from root cron every 30 minutes, but the value got only an emptiness check and a trailing-slash strip. **No privilege boundary is crossed** — the setup already requires root and the value is the admin's own `--portal` argument — so this is parity hygiene, not an RCE. Both setups now use the same character class `ob-builder` enforces via `is_valid_url()`. |
| **B5** | `ob-cert@` / `ob-record@` gained `UMask=0077` and `SystemCallFilter=@system-service` (both run `User=root` by design, so this is defence in depth; the sink `fchmod()`s dirs to 0750 and files to 0640 explicitly, so `UMask` costs auditors nothing). `ob-session-monitor` gained `ProtectSystem=full` — see the correction below. |

### One correction to the B5 premise

`ProtectSystem=full` does **not** work with the `ReadWritePaths` `ob-session-monitor` had. Two independent reasons:

1. The monitor appends to `/etc/open-bastion/force_online` with `>>`, which **creates** the file when missing, and creating an entry needs the **parent directory** writable. A per-file exception under a read-only `/etc` yields `EROFS` — reproduced with the same bind-mount primitives systemd uses, in a user+mount namespace.
2. systemd refuses to set up the namespace at all when a `ReadWritePaths=` entry does not exist. Since `ProtectHome=yes` already forced a namespace, naming the file meant the unit failed to start on **every host that had never gone offline** — a pre-existing bug this change also fixes.

The entry is therefore the directory now, with `openbastion.conf` kept read-only via `ReadOnlyPaths` (applied after `ReadWritePaths`, so the more specific entry wins). Caveat worth knowing: `auth_cache_force_online` is configurable, so a site that points it outside `/etc/open-bastion` will need its own `ReadWritePaths` override.

## C — demo hygiene (#201), commit 3

| | |
|---|---|
| **C1** | All eight demo entrypoints used a fixed `/tmp/admin_cookies`, created world-readable by `touch`, left on disk, and `cat` into the container log on login failure — printing a usable admin portal session cookie. Now `mktemp` + `chmod 600` + `trap ... EXIT`, the pattern `quick-start/server/entrypoint.sh` already had, which also covers the early `exit 1` paths that previously leaked it. The failure branch says the cookie was withheld. |
| **C2** | `docker-demo-cert` and `docker-demo-token` chmod'd the recordings tree `1777` (world-writable — the exact opposite of the product); maxsec and token-svc used `1770`. All four bastion demos now create `ob-sessions` and apply production's `root:ob-sessions 0750` (`debian/open-bastion.postinst`, `src/ob-record-sink.c`), so the demo shows the real tamper-evident layout (#151). |
| **C3** | **Not actionable in this repository.** `local/` is listed in `.gitignore:45`, so `local/ansible/roles/open-bastion/files/lab-curl-shim.sh` and the ansible role around it are untracked lab-only files with no presence in git history — there is nothing to delete in a PR. The tracked `local-test/sso/ob-builder-curl-shim.sh` is a *different* file and is still referenced by `deploy-ansible.sh`, `deploy-shell.sh` and `local-test/README.md`, so it was deliberately left alone. |

## Verification

Clean configure + build, zero compiler warnings or errors:

```
-- Configuring done (0.5s)
-- Generating done (0.0s)

$ cmake --build build-verify -j$(nproc) 2>&1 | grep -iE "warning|error"
(no output)

...

Total Test time (real) =   8.90 sec
```

New test coverage:

```
$ ./build-verify/tests/test_audit_log
  Testing log_created_with_0640... PASS
  Testing log_preserves_stricter_mode... PASS
  Testing module_field_name... PASS
15/15 tests passed

$ bash tests/test_ob_bastion_setup.sh
  PASS: portal URL with shell metacharacters is rejected
  PASS: ordinary portal URLs pass validation
  PASS: sudoers drop-in is visudo-validated before install
  PASS: generated sudoers rule parses (skipped: no visudo)
=== Results: 22/22 passed, 0 failed ===
```

Shell and packaging checks:

- `shellcheck --severity=warning` clean on every script under `scripts/` and on all eight touched demo entrypoints. The remaining `SC2034` notices (`LOGIN_RESP`, the retry loop variable) are present on `main` and unrelated.
- `bash -n` clean on `scripts/ob-*` (the CI step) and on every touched entrypoint.
- `sh -n debian/open-bastion.postinst` OK.
- All 14 `tests/test_ob_*.sh` pass (`test_ob_record_sink.sh` self-skips without a `build/` tree).
- `systemd-analyze verify` on the three changed units: no directive errors, only the expected "binary not installed" notes on this dev host. `systemd-analyze security --offline` confirms the new directives are recognised. **Nothing was restarted on the host.**
- `cmake --install` into a `DESTDIR` confirms the logrotate template lands at `share/open-bastion/logrotate/open-bastion`, matching the new `debian/open-bastion.install` and `%files` entries.


src/audit_log.c
```diff
@@ -302,7 +323,32 @@ int audit_log_event(audit_context_t *ctx, const audit_event_t *event)
             }
```
guimard, line 323: [Revue — Mineur] Le `fchmod` conditionnel ne resserre plus un audit.json préexistant trop permissif : `permissions_ok` ne refuse que world/group-writable, donc un 0644 préexistant (umask laxiste ou outil tiers) est accepté et restera world-readable pour toujours → noms d'utilisateurs, IP, commandes lisibles par tous les comptes locaux. Suggestion : resserrer uniquement dans le sens strict — si `(st.st_mode & 0044) != 0`, `fchmod(fd, st.st_mode & ~0044)` — préserve un 0600 volontaire tout en réparant un 0644.

guimard (commented):
Revue de code — verdict : **APPROUVÉ** (2 points de vigilance, pas de bloquant).

Excellents points : CURLOPT_NOSIGNAL (élimine le hijack SIGALRM dans les processus hôtes multi-thread), db_reset au lieu de db_set "" (secret debconf réellement oublié, y compris Mode=none), readopt_token_file_identity corrige l'auto-lockout après refresh on-demand, verdict négatif /pam/authorize NULL-safe chez le caller, sudoers validé par visudo -cf, cookies de démo mktemp/0600/trap, piège #DEBHELPER# respecté, réarmement des sockets après upgrade préservé côté Debian et mirroiré côté RPM.

Vigilance :
1. (Mineur, détail inline) fchmod conditionnel sur audit.json — voir inline.
2. (Mineur) Durcissement systemd non exercé par la CI (les conteneurs de démo n'ont pas systemd) : UMask=0077 / SystemCallFilter=@system-service sur ob-cert@/ob-record@, ProtectSystem=full + ReadWritePaths sur ob-session-monitor. Un EPERM sur un syscall nécessaire serait amplifié par le fail-closed du recording (sink en échec = logins refusés) → prévoir un test e2e manuel sur hôte réel/VM avant la release.
3. (Nit) La regex de validation d'URL laisse passer un backslash (littéral dans une bracket ERE) — sans conséquence réelle ; et si `auth_cache_force_online` est configuré hors /etc/open-bastion, l'écriture est EROFS silencieuse sous ProtectSystem=full (retour du `>>` non vérifié) — cas custom uniquement.

PR #228, opened 2026-09-05
fix(pam): enforce the SSH key policy fail-closed instead of skipping it

### LINKED ISSUE #181: SSH key policy is enforced nowhere: fail-open on unknown algorithm, ssh_key_min_rsa_bits has no caller (R-S11)
The SSH key policy is not enforced anywhere, on either side of the chain, while `doc/security/02-ssh-connection.md:686-710` lists the R-S11 remediation as "IMPLEMENTEE".

## 1. The PAM-side check is skipped whenever the algorithm is unknown

`src/pam_openbastion.c:3102-3145`:

```c
if (data->ssh_key_policy.enabled &&
    (strcmp(service, "sshd") == 0 || strcmp(service, "ssh") == 0)) {
    char *ssh_algorithm = extract_ssh_algorithm(pamh);
    if (ssh_algorithm) {          /* NULL => whole policy block skipped */
```

`extract_ssh_algorithm()` (`src/pam_openbastion.c:1892-1897`) reads **only** `pam_getenv(pamh, "SSH_USER_AUTH")`. It never consults `read_ssh_fp_from_spool()`, the out-of-band channel that exists precisely because sshd does not export `SSH_USER_AUTH` during `pam_acct_mgmt` on OpenSSH >= 9.8 — a limitation the project documents itself at `doc/pam-modes.md:241-258`.

So on any current OpenSSH the policy silently does nothing instead of failing closed. The spool file only carries `SHA256:<fp>` and not the algorithm, so the check cannot currently be rescued from it either.

## 2. The minimum RSA size is dead code

`src/ssh_key_policy.c:247-256`, RSA branch:

```c
/* For RSA, we can't check the key size from the algorithm alone. */
local_result.valid = true;
```

`ssh_key_policy_check_rsa_size()` (`src/ssh_key_policy.c:209`) has **no production caller**: a repo-wide grep returns 3 hits, all in `tests/test_ssh_key_policy.c:407-409`. The same holds for `ssh_key_policy_check()` — `grep -rn ssh_key_policy_check --include=*.c .` matches only the test file.

## 3. The CA does not enforce a policy either

The ssh-ca plugin signs whatever it is given after a purely syntactic regex check (`SSHCA.pm:282-286`), with no key-type allowlist, no minimum bit length and no blob decode. Verified on OpenSSH 10.4p1: an RSA-1024 key signs successfully. Filed separately in `linagora/lemonldap-ng-plugins`. Note that DSA and RSA-512 are refused by modern `ssh-keygen` itself, so RSA-1024 is the realistic weak case.

## Consequences

`ssh_key_min_rsa_bits` is inert configuration, and R-S11 is claimed as remediated in the EBIOS study on the strength of a control that runs nowhere.

## Suggested fix

Either make the control real — carry the algorithm through the fingerprint spool, call `ssh_key_policy_check_rsa_size()`, and fail closed when the algorithm cannot be determined while the policy is enabled — or remove the dead code and correct the R-S11 claim in `doc/security/`.

Precondition for (1): `ssh_key_policy_enabled=true`, which is not the default (`src/config.c:170`).

---

guimard:
Fixes #181

## The problem: the policy was enforced nowhere

`doc/security/02-ssh-connection.md` listed the R-S11 remediation as
`IMPLÉMENTÉE`, but on any current OpenSSH `ssh_key_policy_enabled = true`
enforced exactly nothing.

1. **Fail-open on the only path that matters.** `src/pam_openbastion.c` did:

   ```c
   char *ssh_algorithm = extract_ssh_algorithm(pamh);
   if (ssh_algorithm) {          /* NULL => whole policy block skipped */
   ```

2. **And NULL was the normal case.** `extract_ssh_algorithm()` reads only
   `pam_getenv(pamh, "SSH_USER_AUTH")`, which sshd does not export to the PAM
   environment during `pam_acct_mgmt` on OpenSSH >= 9.8 — the project documents
   this itself in `doc/pam-modes.md`. So the block was skipped on every login.

3. **`ssh_key_min_rsa_bits` was inert.** The RSA branch of
   `ssh_key_policy_check()` set `valid = true` with the comment *"For RSA, we
   can't check the key size from the algorithm alone"*, and
   `ssh_key_policy_check_rsa_size()` had no production caller —
   `grep -rn ssh_key_policy_check --include=*.c .` matched only the test file.

The decision was to implement it fail-closed.

### Carries the key out of band (spool extension)

`AuthorizedPrincipalsCommand` accepts `%t` (key/certificate type) and `%k`
(base64 key or certificate blob) — both since **OpenSSH 7.4**, so well below
every supported target. `ob-ssh-principals`, generated by `ob-bastion-setup` /
`ob-backend-setup`, is now called as `... %u %f %t %k` (`... %u %f %i %t %k` on
the backend) and writes a **second** drop next to the existing one:

```
/run/open-bastion/ssh-fp/<anchor>.fp    unchanged: "SHA256:<base64>"
/run/open-bastion/ssh-fp/<anchor>.key   v=1 / fp= / alg= / key=
```

Two files rather than a changed format, on purpose: the `.fp` drop keeps its
exact previous bytes, so an **older `pam_openbastion` next to a newer helper**
keeps working byte-for-byte. Unknown keys in the `.key` file are ignored, so the
format can grow. Both drops go through the same `read_spool_drop()` and keep all
the existing integrity checks — spool dir not group/world-writable, regular file
owned by the spool-dir owner, `nlink == 1`, mode `0600`, size-capped. The helper
filters `%t` and `%k` against a strict character set before spooling.

### Decodes the key rather than trusting a type string

`ssh_key_blob_info()` (in `src/ssh_key_policy.c`) base64-decodes the blob and
walks the SSH wire format (RFC 4253 §6.6 / `PROTOCOL.certkeys`), skipping the
certificate nonce, to get the real type and the real size — including the **RSA
modulus size**, which exists nowhere else. This mirrors what the ssh-ca plugin
PR #78 does in Perl (`_parseSshPubKey` / `_mpintBits`). The module cross-checks
the metadata's `fp=` against the `.fp` drop and discards a stale one.

### Fails closed, and wires up the RSA size check

- New `ssh_key_policy_check_key(policy, algorithm, key_bits, result)` is the
  enforcing entry point: it calls `ssh_key_policy_check_rsa_size()`, and rejects
  an RSA key whose size could not be measured.
- `ssh_key_policy_check()` remains as the type-only variant, now documented as
  such — and a test pins the difference so nobody re-wires production to it.
- When the policy is enabled and the key cannot be identified at all (no `.key`
  drop and no `SSH_USER_AUTH`), `pam_sm_acct_mgmt` returns `PAM_PERM_DENIED`,
  logs at ERR naming the fix, and records an `AUDIT_AUTHZ_DENIED` event.

## Upgrade order (deliberately handled)

`ssh_key_policy_enabled` defaults to `false` (`src/config.c:170`) and the block
is unreachable while it is off, so **the default deployment is untouched** —
`check_key_disabled_policy_unchanged` pins that.

The risky combination is a host that *had* enabled the (inert) policy: `apt
upgrade` replaces the PAM module but **not** the helper, which lives in
`/usr/local/sbin` and is written by the setup script, not shipped by the
package. A new module + old helper would deny every SSH login. Handled three
ways:

1. The helper carries a `spool-format: v1` marker, and the **postinst** warns
   loudly when it finds `ssh_key_policy_enabled` set in `openbastion.conf` next
   to a pre-v1 helper, naming the setup script to re-run. Hosts with the policy
   off print nothing.
2. The docs state the precondition and give the one-line check:
   `grep -q 'spool-format: v1' /usr/local/sbin/ob-ssh-principals && echo OK`.
3. The reverse order is safe by construction: a new helper called by an old
   `sshd_config` (2 arguments, no `%t`/`%k`) still writes `.fp` and simply
   writes no `.key` — covered by a test.

## Docs

`doc/security/02-ssh-connection.md` (R-S11 now says implemented **with its
precondition**, and the fail-closed behaviour), `doc/pam-modes.md` (the capture
section rewritten for the two drops, plus a new "SSH key policy enforcement"
subsection), `doc/security.md`, `doc/admin-guide.md` (including real error
strings in Troubleshooting), `SECURITY.md`, `config/openbastion.conf.example`,
`CHANGELOG.md`.

## Verification

Baseline before the change was a clean build and 19/19.

```
-- Configuring done (0.2s)
-- Generating done (0.0s)
-- Build files have been written to: .../build-verify

$ cmake --build build-verify -j$(nproc)
[100%] Built target test_notify
[100%] Built target builder-deb
   (0 warnings, 0 errors; the project builds with -Wall -Wextra -Werror)

 ...

Total Test time (real) =   8.87 sec
```

`tests/test_ssh_key_policy.c` goes from 33 to 47 tests:

```
$ ./build-verify/tests/test_ssh_key_policy
...
Running blob_info_rsa_2048... PASSED
Running blob_info_rsa_1024... PASSED
Running blob_info_ed25519... PASSED
Running blob_info_ecdsa... PASSED
Running blob_info_certificate... PASSED
Running blob_info_rejects_garbage... PASSED
Running blob_info_rejects_oversized... PASSED
Running check_key_rsa_size_enforced... PASSED
Running check_key_rsa_unknown_size_fails_closed... PASSED
Running check_key_fixed_size_types... PASSED
Running check_key_disabled_policy_unchanged... PASSED
Running check_key_no_algorithm_denied... PASSED
Running check_type_only_does_not_enforce_rsa_size... PASSED
Running scenario_blob_to_policy_decision... PASSED

====================
Tests: 47 passed, 0 failed
```

The blobs are real `ssh-keygen` output (RSA-2048, RSA-1024, Ed25519, ECDSA
P-256/P-521, and RSA/Ed25519 certificates), so the decoder is checked against
sizes OpenSSH itself reports.

**Mutation check** — flipping the new enforcement flag back off makes exactly
the new guards fail, so they are not vacuous:

```
Running check_key_rsa_size_enforced... FAILED at tests/test_ssh_key_policy.c:667: !ssh_key_policy_check_key(&policy, "ssh-rsa", 1024, &result)
Running check_key_rsa_unknown_size_fails_closed... FAILED at tests/test_ssh_key_policy.c:697: !ssh_key_policy_check_key(&policy, "ssh-rsa", 0, &result)
Running check_type_only_does_not_enforce_rsa_size... FAILED at tests/test_ssh_key_policy.c:775: !ssh_key_policy_check_key(&policy, "ssh-rsa", 2048, &result)
Running scenario_blob_to_policy_decision... FAILED at tests/test_ssh_key_policy.c:793: !ssh_key_policy_check_key(&policy, algo, bits, &result)
Tests: 47 passed, 4 failed
```

New shell test (picked up by CI's `tests/test_ob_*.sh` loop); it extracts both
helpers from the setup scripts so there is a single source of truth, and runs
them under a process named `sshd-session` so the anchor walk behaves as it does
under a real sshd:

```
$ bash tests/test_ob_ssh_principals_spool.sh
=== ob-ssh-principals spool tests ===
  PASS: Helpers extracted and syntactically valid
  PASS: Bastion helper writes .fp (unchanged) and v=1 .key drop
  PASS: Spool drops are mode 0600
  PASS: Malformed key type is filtered (no .key drop, .fp still written)
  PASS: Malformed key blob is dropped, algorithm still recorded
  PASS: Legacy 2-argument invocation still writes .fp and no .key
  PASS: Backend helper: vouched cert accepted and both drops written
  PASS: Backend helper still denies an unvouched (direct SSO) cert
  PASS: Backend helper still denies a bastion_id outside the allowlist
  PASS: sshd_config templates pass %t and %k to the helper
  PASS: Helpers advertise 'spool-format: v1' for the postinst upgrade guard

Tests run: 11, passed: 11, failed: 0
```

No regression in the rest of the shell suite:

```
tests/test_ob_backend_setup.sh                   rc=0  === Results: 18/18 passed, 0 failed ===
tests/test_ob_bastion_setup_audit.sh             rc=0  === Results: 11/11 passed, 0 failed ===
tests/test_ob_bastion_setup_hardening.sh         rc=0  === Results: 20/20 passed, 0 failed ===
tests/test_ob_bastion_setup.sh                   rc=0  === Results: 18/18 passed, 0 failed ===
tests/test_ob_builder_service_accounts.sh        rc=0  All tests passed!
tests/test_ob_enroll.sh                          rc=0  All tests passed!
tests/test_ob_heartbeat.sh                       rc=0  === Results: 22/22 passed, 0 failed ===
tests/test_ob_record_sink.sh                     rc=0
tests/test_ob_scp.sh                             rc=0  All tests passed!
tests/test_ob_session_prune.sh                   rc=0  === Results: 10/10 passed, 0 failed ===
tests/test_ob_session_recorder.sh                rc=0  === Results: 16/16 passed, 0 failed ===
tests/test_ob_sftp.sh                            rc=0  All tests passed!
tests/test_ob_ssh_cert.sh                        rc=0  === Results: 17/17 passed, 0 failed ===
tests/test_ob_sshd_dropin_precedence.sh          rc=0
tests/test_ob_ssh_principals_spool.sh            rc=0  Tests run: 11, passed: 11, failed: 0
tests/test_ob_ssh.sh                             rc=0  All tests passed!
```

Lint on every shell touched, including the helpers extracted out of their
heredocs and the **assembled** postinst (the `#DEBHELPER#`-in-a-comment trap
from 0.6.1):

```
$ bash -n scripts/ob-bastion-setup scripts/ob-backend-setup   # OK
$ shellcheck scripts/ob-bastion-setup scripts/ob-backend-setup tests/test_ob_ssh_principals_spool.sh   # clean
$ sh -n / shellcheck -s sh <extracted ob-ssh-principals, both roles>   # clean
$ sed 's/#DEBHELPER#/.../' debian/open-bastion.postinst | sh -n   # OK, and the token appears exactly once, not in a comment
```

The new decoder was also run under ASan + UBSan: the 47 unit tests pass clean,
and a 600 000-iteration randomized smoke test over base64-shaped inputs and
mutations of a real blob produced no crash and no sanitizer report.

Markdown reformatted with `prettier` (the two files it touched were clean
before, and are clean after).


src/pam_openbastion.c
```diff
@@ -1680,24 +1680,41 @@ static pid_t find_sshd_session_ancestor(void)
...
  * channel that compensates for the fact that OpenSSH does not propagate
  * SSH_USER_AUTH to the PAM environment during pam_acct_mgmt.
  *
+ * Two suffixes exist:
+ *   ".fp"  - v0 format, a bare "SHA256:<base64>" line (fingerprint binding)
+ *   ".key" - v1 format, "k=v" lines describing the presented key (key policy)
+ * They are separate files on purpose: an older pam_openbastion keeps reading
+ * ".fp" byte-for-byte as before, so a newer helper never breaks it.
+ *
  * The file is intentionally left in place on successful read: the same
```
guimard, line 1698: [Revue — Mineur] Le spool `/run/open-bastion/ssh-fp` n'est jamais purgé et cette PR double le nombre de fichiers par connexion (`.fp` + `.key`, jusqu'à ~20 Ko). Le tmpfiles (`d /run/open-bastion/ssh-fp 0700 ...` dans ob-bastion-setup:376 / ob-backend-setup:418) crée sans âge de purge, et aucun `unlink` post-lecture. Avec `kernel.pid_max` élevé les PID ne rebouclent pas avant des mois → ~35 Mo/jour dans `/run` pour ~10k connexions, jusqu'à remplir le tmpfs ; `mktemp` du helper échoue alors → toutes les connexions refusées (fail-closed). Suggestion : purge par âge dans tmpfiles (type `e`, ex. `1h` — le helper réécrit les drops à chaque connexion) ou `unlink` des deux drops dans `pam_sm_close_session` (les drops doivent persister pendant la session pour sudo `/pam/verify`, la fin de session est le bon moment).

debian/open-bastion.postinst
```diff
@@ -297,4 +297,31 @@ if [ "$1" = configure ] \
...
+# type and the key blob reach the module through the v1 key-metadata drop
+# written by /usr/local/sbin/ob-ssh-principals — and that helper is (re)written
+# only by ob-bastion-setup / ob-backend-setup, never by this package. A plain
+# upgrade therefore leaves a NEW module next to an OLD helper. Warn loudly here
+# instead of letting the admin find out at the next login. Hosts that never
+# enabled the policy (the default) are not concerned and print nothing.
+_ob_conf=/etc/open-bastion/openbastion.conf
+_ob_helper=/usr/local/sbin/ob-ssh-principals
+if [ "$1" = configure ] && [ -f "$_ob_conf" ] && [ -f "$_ob_helper" ] \
+    && grep -Eq '^[[:space:]]*ssh_key_policy(_enabled)?[[:space:]]*=[[:space:]]*(1|true|yes|on)[[:space:]]*$' "$_ob_conf" \
```
guimard, line 323: [Revue — Faible] La regex de détection de `ssh_key_policy_enabled` ne matche pas une valeur citée (`= "true"`), alors que le parseur C retire les guillemets et l'active (`src/config.c` ~755-763). Admin avec style INI + helper pré-v1 après `apt upgrade` → lockout de tous les logins sans l'avertissement prévu — exactement le cas que ce postinst est censé attraper. Suggestion : normaliser la ligne (`sed` supprimant `["']` autour de la valeur) avant le grep, ou documenter la limite.

guimard (commented):
Revue de code — verdict : **APPROUVÉ AVEC RÉSERVES** (aucun bloquant).

Le fix ferme réellement #181 sans régression sur les chemins d'autorisation LLNG : aucun appel /pam/authorize ni résolution de groupe de serveurs touché, compat binaire du `.fp` préservée et testée, décodeur de blob bien défendu (ASan/UBSan), postinst assemblé vérifié (`sh -n`), CI 16/16 verte.

Réserves (détail inline) :
1. Spool `/run/open-bastion/ssh-fp` jamais purgé (Mineur — risque d'auto-DoS).
2. Postinst ne détecte pas `ssh_key_policy_enabled = "true"` cité (Faible).
3. TOCTOU validation/unlink du drop (Faible).
4. Chemin C spool→refus sans test automatisé (Faible).

Note : la politique s'applique aussi aux comptes de service (comportement correct, mais non documenté — une phrase dans doc/security.md / doc/admin-guide.md éviterait de casser des jobs d'automatisation par surprise).

src/pam_openbastion.c
```diff
@@ -1729,19 +1746,19 @@ static char *read_ssh_fp_from_spool(pam_handle_t *pamh)
     }
 
     char path[128];
-    int n = snprintf(path, sizeof(path), "%s/%d.fp",
-                     OB_SSH_FP_SPOOL_DIR, (int)anchor);
+    int n = snprintf(path, sizeof(path), "%s/%d.%s",
+                     OB_SSH_FP_SPOOL_DIR, (int)anchor, suffix);
     if (n < 0 || (size_t)n >= sizeof(path)) return NULL;
 
     int fd = open(path, O_RDONLY | O_NOFOLLOW | O_CLOEXEC);
```
guimard, line 1757: [Revue — Faible] TOCTOU : le drop est validé via le descripteur ouvert (`O_NOFOLLOW`, `fstat`, `nlink==1`) mais le `unlink` est fait par chemin ensuite. Une connexion B peut remplacer le fichier (`mv -f` du helper) entre lecture et `unlink` → suppression du drop frais de B (deny transitoire pour une connexion légitime, pas d'escalade). Motif pré-existant pour `.fp`, étendu ici à `.key`. Suggestion : `fstat` sur le chemin avant `unlink` et comparer `st_dev`/`st_ino`, ou `rename()` vers un nom `.stale`.

src/pam_openbastion.c
```diff
@@ -3095,52 +3298,102 @@ PAM_VISIBLE PAM_EXTERN int pam_sm_acct_mgmt(pam_handle_t *pamh,
...
+     * environment during pam_acct_mgmt.
+     *
+     * The key is identified from the ".key" drop written by ob-ssh-principals
+     * (sshd's %t and %k tokens), with SSH_USER_AUTH kept as a fallback for
+     * sshd variants that do export it.
+     *
+     * Nothing here runs unless the policy is explicitly enabled
+     * (ssh_key_policy_enabled defaults to false).
      */
     if (data->ssh_key_policy.enabled &&
```
guimard, line 3324: [Revue — Faible] La branche la plus dangereuse (`read_ssh_key_meta_from_spool` → `have_meta=false` → `PAM_PERM_DENIED` + log) n'est couverte par aucun test automatisé : `tests/test_ssh_key_policy.c` n'exerce que les fonctions pures, aucun test d'intégration Docker n'active la politique, et le grep du postinst n'est pas testé. Suggestion : test d'intégration `pamtester`/`pam.d` factice avec politique activée et spool vide (attendu : `PAM_PERM_DENIED` + log) + test shell du postinst (config activée + helper sans marqueur → avertissement ; avec marqueur → silence).

PR #224, opened 2026-09-05
refactor: remove dead token cache, client_context, and keyring settings

guimard:
Fixes #193

`SECURITY.md` documented three things as active security features that are not wired into the PAM chain at all. Following the decision in the issue, this deletes the dead code and corrects the documentation rather than rewiring it — rewiring the token cache would reopen an offline authentication path nothing currently needs, and dead code is attack surface with no compensating benefit.

**This is a documentation-accuracy fix as much as a cleanup.** The docs claimed security properties (encryption at rest, risk-based TTL, kernel-keyring storage) that the running module did not have.

## What I verified as unreferenced

Grepped the entire repo — `src/`, `nss/`, `include/`, `tests/`, `scripts/`, `admin-builder/`, `debian/`, `rpm/`, `lightdm/`, `systemd/`, `config/`, `man/`, `doc/`, `.github/` and both `CMakeLists.txt` — **including the `ENABLE_DESKTOP_SSO` paths**, which do ship (`debian/rules` and `rpm/open-bastion.spec` both build with `-DINSTALL_DESKTOP=ON`). Both build configurations were compiled and tested.

| Item | Finding |
| --- | --- |
| `cache_lookup()` / `cache_store()` | Zero callers outside `tests/test_token_cache.c`. Nothing ever wrote a token into the cache or read one back. |
| `data->cache` in `pam_openbastion.c` | Only init, destroy, and `cache_invalidate_user()` — invalidating a cache nothing filled. |
| `client_context_*` (all 6 entry points) | Zero callers outside `tests/test_client_context.c`. |
| `add_key` / `request_key` / `keyctl` | **Do not appear anywhere in the tree.** `secret_store` only `strdup`'d and `free`'d `use_keyring` / `keyring_name`. |
| `secret_store_init()` | Note: has **zero callers in `src/`** either — the whole module is only exercised by its unit test. Left in place; out of scope for this issue. |
| lightdm / greeter | Pure JS/CSS/HTML + a `.conf`; no reference to any removed symbol. |
| `nss/libnss_openbastion.c` | Has its **own** config parser reading `/etc/open-bastion/nss_openbastion.conf` with its own `cache_ttl`. Untouched. |

Two same-named-but-unrelated things I checked and did **not** touch:
- `src/crowdsec.c` defines its own `static cache_lookup()` / `cache_store()` for CrowdSec decisions — live code, different symbols.
[...]

PR #232, opened 2026-09-06
refactor: remove secret_store, the last dead cryptographic module

### LINKED ISSUE #225: secret_store is compiled in but has no callers — dead code like #193
Found while removing the other dead cache code (#193). The `secret_store` module is compiled into the PAM module but nothing calls it.

## Evidence

`CMakeLists.txt:42` builds `src/secret_store.c` into the module. But:

```
$ grep -rn "secret_store_" --include=*.c --include=*.h src/ nss/ include/ tests/ \
    | grep -v '^src/secret_store.c' | grep -v '^include/secret_store.h' | grep -v '^tests/test_secret_store.c'
(no output)
```

Every entry point — `secret_store_init`, `secret_store_get`, `secret_store_put` and the rest — is reachable only from `tests/test_secret_store.c`.

Consequently the `secrets_encrypted` config key (`src/config.c:118,513-514,841`) is parsed, defaulted and validated but never reaches any store, and `secrets_use_keyring` / `keyring_name` were already unimplemented (removed in #193).

## Why this matters now

Two recently filed issues land inside this dead module and should be re-read in that light:

* **#187** — `secret_store_get` decrypting into the caller buffer without a size check. Fixed in PR #222, which is right regardless, but the "defence in depth" framing was generous: there is no caller to defend.
* **#184** — the note that `secret_store` derives its key from `/etc/machine-id` only, with no key-file support, and that `SECURITY.md:660`'s "machine/key binding" claim overstates it. If the module is dead, the honest fix is to stop claiming the capability rather than to strengthen it.

## Decision needed

Same question as #193, and the answer there was to delete:

* **Delete** the module, its config keys, its test and the `SECURITY.md` claims that describe it. Removes dead attack surface compiled into a root PAM module.
* **Or wire it up**, if there is a planned consumer — in which case say what it is, because the encryption story then needs the key-file work from #184 first.

Deleting is the default recommendation, consistent with #193.

---
*Reported from a security/conformity audit pass; verified against `main` at the time of filing.*

guimard:
Closes #225.

Same shape as the dead code removed in #224. `src/secret_store.c` was compiled into the PAM module and every entry point — `secret_store_init`, `_get`, `_put`, `_delete`, `_rotate_key` — was reachable only from its own unit test. Nothing in the authentication path ever stored or read a secret through it.

## What is removed

- `src/secret_store.c`, `include/secret_store.h`, `tests/test_secret_store.c`, and their build entries.
- The `secrets_encrypted` config key and the `no_encrypt_secrets` module argument, parsed, defaulted and validated without ever reaching a store.

This takes AES-GCM code and file writes under `/etc/open-bastion` out of a root PAM module that had no use for them.

## Two open findings settle with it

- **#187** — the missing size check in `secret_store_get`. Fixed in #222, correct on its own terms, and now moot: there was no caller to defend.
- **#184** — machine-id-only key derivation with no key-file support, and `SECURITY.md`'s overstated "machine/key binding". The honest fix is to stop claiming the capability, not to strengthen an unused one.

## Documentation

`SECURITY.md` and `doc/security/00-architecture.md` both advertised `secrets_encrypted = true`, "encrypt secrets at rest". They now state what is true: secrets in `openbastion.conf` are protected by file permissions only — root-owned `0600`, enforced by `config_load()`, which refuses to read the file otherwise — and the way to avoid a secret on disk is not to write one (`client_secret_mode: prompt`, or `ansible-vault`).

**Not** removed: `src/cache_key.c` and `src/offline_cache.c`, which do the machine-id/key-file derivation for the offline *credential* cache and are genuinely used. `SECURITY.md`'s cache-binding section describes that one and is unchanged.

## Verification

```
$ grep -rn "secret_store" src/ nss/ include/ tests/ CMakeLists.txt
(no output)
$ ctest --output-on-failure
```

An existing `openbastion.conf` still loads: `secrets_encrypted` is now an unknown key, which is ignored exactly as it was before (and, with #231, logged).


guimard:
Revue du 06/09/2026 — verdict : APPROVE (2 nits)

Retrait propre : aucun résidu de secret_store (code/headers/build/tests/packaging/configs exemples), build Release + ctest 19/19 rejoués, client_secret_jwt / ob-bastion-id / ob-builder --output intacts, rien n'était installé (aucun changement debian/rpm nécessaire). Les docs réécrites disent la vérité (vérifié dans src/config.c).

- [nit] `SECURITY.md:437-442` : « (and, with #231, logged) » est prospectif tant que #231 n'est pas mergée — après fusion, les confs contenant `secrets_encrypted = true` émettront un warning « unknown configuration key ». À noter à l'ordre de merge.
- [nit] `SECURITY.md:433` : le code vérifie les bits rw groupe/autres (`src/config.c:43`), pas les bits d'exécution, et n'exige pas 0600 stricte côté propriétaire — formulation légèrement plus forte que le code.

PR #234, opened 2026-09-06
fix(cert): drop the privileged fallback and let the host-key policy be tightened

### LINKED ISSUE #202: Document or replace StrictHostKeyChecking=accept-new; remove the root fallback in ob-cert-lib.sh
## 1. TOFU on the first backend hop

`scripts/ob-ssh:105`, `scripts/ob-scp:182` and `scripts/ob-sftp:158` all pass `-o StrictHostKeyChecking=accept-new`. The first connection to any backend trusts whatever host key is presented, so an attacker on the bastion-to-backend path can MITM that first hop and read session content. Subsequent connections are pinned by `known_hosts`.

The ephemeral private key never leaves the bastion and the certificate is source-address-pinned, so credentials are not stolen. This is a deliberate usability trade-off, but it does not appear to be documented as a limitation anywhere.

## 2. Root/lab fallback in `ob-cert-lib.sh` bypasses the daemon

`scripts/ob-cert-lib.sh:167-172`:

```sh
if [ -r "$SERVER_TOKEN_FILE" ]; then
    # Privileged path (root, or a relaxed lab setup): call LLNG directly.
    server_token=$(get_server_token)
    ... -H "Authorization: Bearer $server_token" ...
```

The guard is `[ -r ]` on a file that `scripts/ob-bastion-setup:1511` sets to 0600 root, so the branch is only taken when the caller is already root or the lab has deliberately relaxed permissions. It is not reachable by an unprivileged user on a correctly deployed bastion, and the token goes through a shell variable rather than argv so `/proc/<pid>/cmdline` does not expose it. Still an untidy escape hatch around the SO_PEERCRED daemon design.

## Suggested fix

Document the TOFU trade-off (or offer a pre-seeded `known_hosts` path), and remove the privileged fallback branch so the server token path is always the daemon.

---

guimard:
Closes #202 — both halves.

## 1. The privileged fallback is gone

`request_bastion_cert()` took a `[ -r "$SERVER_TOKEN_FILE" ]` branch that called `/pam/bastion-cert` directly with the bastion's bearer token whenever the caller could read the token file — root, or a lab that had relaxed the `0600`. An escape hatch around the SO_PEERCRED design, with none of the daemon's checks.

Root and unprivileged callers now take the same audited path through `ob-cert-request` → `ob-cert-daemon`. `get_server_token()` and `build_curl_opts()` went with it: the shared library no longer contacts the portal at all.

One correction to the issue's reading: `SERVER_TOKEN_FILE` is **not** dead config. `ob-cert-daemon` parses the same `ssh-proxy.conf` and reads the token path from it (`src/ob-cert-daemon.c:229`). The key stays; only the shell library stops using it.

## 2. The TOFU default could not be overridden — that was the real bug

The three connectors passed `-o StrictHostKeyChecking=accept-new` **before** the operator's `SSH_OPTIONS`. `ssh` keeps the **first** value it is given for an option, so a site that pre-seeded `known_hosts` had no way to refuse an unknown backend host key — the documented `SSH_OPTIONS` escape hatch silently did nothing for this option.

The default is now emitted only when `SSH_OPTIONS` does not set `StrictHostKeyChecking` itself (both the `-o X=y` and `-oX=y` spellings), so this works:

```
SSH_OPTIONS="-o StrictHostKeyChecking=yes -o GlobalKnownHostsFile=/etc/ssh/ssh_known_hosts"
```

The trade-off itself was documented nowhere, as the issue says. It now appears in `doc/admin-guide.md`, `man ob-ssh` and risk **R-S9** of the EBIOS study, each stating precisely what an attacker on the bastion→backend path gets (the first session's content) and what they do not (anything reusable — the ephemeral key never leaves the bastion, the certificate is source-address pinned), with the pre-seeded `known_hosts` remediation.

## Tests

`tests/test_ob_ssh.sh` loses the `get_server_token` / `build_curl_opts` cases (the functions no longer exist) and gains four: one asserting the library has no direct portal path left, and three on the host-key policy — default, operator override, joined `-o` form.

```
$ bash tests/test_ob_ssh.sh   → 22/22
$ bash tests/test_ob_scp.sh   → 7/7
$ bash tests/test_ob_sftp.sh  → 6/6
$ shellcheck -S warning scripts/ob-cert-lib.sh scripts/ob-{ssh,scp,sftp}  → clean
```

## Blast radius worth knowing

A container demo where `ob-ssh` is run as **root** and no `ob-cert.socket` exists used to work through the fallback and now fails with the daemon's message. The docker demos ship the token `0600`, so their documented `dwho` workflow already required the daemon; no integration test exercises `ob-ssh`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   so there is nothing to hide. The repository's own `.gitignore` also covers the
   bundle directories that were sitting untracked in the working tree.
 
+- **`ob-ssh` / `ob-scp` / `ob-sftp` no longer have a privileged shortcut around
+  `ob-cert-daemon` (#202).** `request_bastion_cert()` took a
+  `[ -r "$SERVER_TOKEN_FILE" ]` branch that called `/pam/bastion-cert` directly
+  with the bastion's bearer token whenever the caller could read the token file
+  — root, or a lab that had relaxed the `0600`. It was an escape hatch around
+  the SO_PEERCRED design, with none of the daemon's checks. It is gone: root and
+  unprivileged callers now take the same audited path through
+  `ob-cert-request` → `ob-cert-daemon`. `get_server_token()` and
+  `build_curl_opts()` went with it — the shared library no longer contacts the
+  portal at all. `SERVER_TOKEN_FILE` stays in `ssh-proxy.conf`: it is read by
+  `ob-cert-daemon`, which parses the same file.
+
+- **The bastion→backend host-key policy can now be tightened (#202).** The three
+  connectors passed `-o StrictHostKeyChecking=accept-new` *before* the
+  operator's `SSH_OPTIONS`, and `ssh` keeps the **first** value it is given for
+  an option — so the trust-on-first-use default could not be overridden at all.
+  The default is now emitted only when `SSH_OPTIONS` does not set the option, so
+  a site that pre-seeds `/etc/ssh/ssh_known_hosts` can refuse unknown backend
+  host keys:
+
+  ```
+  SSH_OPTIONS="-o StrictHostKeyChecking=yes -o GlobalKnownHostsFile=/etc/ssh/ssh_known_hosts"
+  ```
+
+  The TOFU trade-off itself — an attacker on the bastion→backend path can MITM
+  the *first* connection to a backend and read the session, though nothing
+  reusable is captured — is now documented in `doc/admin-guide.md`,
+  `man ob-ssh` and risk R-S9 of the EBIOS study, where it had never been stated.
+
 - **A world- or group-readable `/etc/open-bastion/cache.key` is now rejected
   instead of used with a warning** (offline auth cache, desktop SSO). Versions
   up to 0.6.2 accepted such a key and merely logged a warning. `SECURITY.md`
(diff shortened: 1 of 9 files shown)


guimard:
Revue du 06/09/2026 — verdict : APPROVE (1 mineur, 4 nits)

Suppression propre du fallback privilégié : aucun chemin résiduel (seuls consommateurs de la lib : ob-ssh/ob-scp/ob-sftp + tests), aucun setuid/sudo réintroduit, chemin unique ob-cert-request → socket → ob-cert-daemon (SO_PEERCRED), token lu uniquement par le démon (open_checked : root-owned, régulier, O_NOFOLLOW + fstat, aucun bit group/other). Host-key policy double garde (détection + ordre « première valeur gagne » de ssh), défaut `accept-new` seulement sans override opérateur, TOFU documenté (admin-guide, man, EBIOS R-S9). Socket inactive → message actionnable. Tests 22/22 + 7/7 + 6/6, shellcheck propre.

- [mineur] `docker-demo-cert/README.md:105,473-476` et `docker-demo-token/README.md:82` décrivent encore « ob-ssh mints an ephemeral certificate via /pam/bastion-cert » — après la PR, ob-ssh ne contacte plus jamais le portail (c'est le démon). Corriger ces descriptions.
- [nit] `man/ob-ssh.1:107` : antécédent de « which » ambigu (c'est ob-cert-daemon qui lit SERVER_TOKEN_FILE, pas ob-cert-request).
- [nit] `scripts/ob-ssh:241` : le `--help` liste SERVER_TOKEN_FILE/TIMEOUT/VERIFY_SSL sans préciser qui les lit (TIMEOUT est d'ailleurs mort dans la lib — parsé mais inutilisé).
- [nit] `scripts/ob-cert-lib.sh:189` : « Until 0.6.2 » → « through 0.6.2 » (la branche existe dans 0.6.2).
- [nit] `tests/test_ob_ssh.sh` : les tests host-key n'assertent pas l'ordre SSH_OPTIONS avant HOST_KEY_OPTS — un test avec ssh stubé renforcerait le filet.

PR #227, opened 2026-09-05
docs: correct documentation drift against the code (#204–#211)

guimard:
Documentation-only correction pass against the code, one commit per issue so they can be reviewed independently. No C source and no setup-script behaviour is changed; the only non-doc edits are help text and comments in `scripts/ob-bastion-id` and `scripts/ob-session-recorder`, plus `man/ob-bastion-id.1`.

## What's in it

| Commit | Issue | Gist |
| --- | --- | --- |
| `7e7625f` | #204 | `AuthorizedPrincipalsCommand /bin/echo %u` in the docs disabled fingerprint binding (`%f`) and the pre-PAM `allowed_bastions` check (`%i`). Corrected in `doc/pam-modes.md` (twice) and `doc/admin-guide.md`, where it was listed under "It performs all of the following automatically". |
| `1222206` | #205 | `SECURITY.md`: supported-versions table (`1.x` → `0.6.x`), CA path, helper path, and the KDF description (the salt is 16 random bytes, not the "cache username"). |
| `d88df1a` | #206 | CA key path, helper path, `ob-heartbeat.timer`, non-existent `libnss-openbastion` / `nss-openbastion` / `lightdm-openbastion-greeter` packages, `ob-standalone-setup` (it exists), a dead anchor link, and the hardening template paths. **Adds `doc/reference-paths.md`** as the single facts page, linked from both indexes and the admin guide. |
| `e22eb07` | #207, refs #178 | The documented Mode E PAM stack now matches the generated one (`pam_mkhomedir`, `session pam_openbastion`, `pam_systemd` added; spurious `account pam_unix` removed), notes `ssh_cert_aware=true` is a no-op, and writes the promised `timestamp_timeout` section. |
| `476a3c4` | #208 | `bastion_id` is a synthetic per-device identity, not the OIDC `client_id`. Also the stale `bastion_jwt_allowed_bastions` help text. |
| `d180022` | #209 | `OB_*` not `LLNG_*` env vars; `.typescript`/`.json` not `.cast`/`.ttyrec`; `doc/audit.md`'s premise was fixed in 0.5.0. |
| `c639f4c` | #210 | auditd shipped in 0.2.0, both design docs marked implemented, and §11 of the tamper-evident design marked "not implemented as designed" with the shipped `pam_systemd` fix. |
[...]

PR #237, opened 2026-09-06
feat(sudo): --enable-sudo-fresh-otp, an LLNG token on every elevation

### LINKED ISSUE #178: sudo timestamp cache lets SSO users sudo without a fresh OTP (Mode E bypass)
## Summary

On a backend (and any host where sudo for SSO users goes through `pam_openbastion`), once a user has done one successful `sudo` with an LLNG one-time token (OTP), they can keep running `sudo`/`sudo -i` **without being asked for a token again**, well beyond a few minutes. This defeats the Mode E guarantee that each sudo elevation requires a fresh LLNG verification.

## Evidence

First sudo runs the full stack (auth = OTP via `/pam/verify`, then account):

```
sudo[199650]: pam_openbastion(sudo:auth): Authenticating user: xguimard
pam_llng:    pam_openbastion(sudo:auth): SSH fp recovered from spool: SHA256:...
pam_llng:    pam_openbastion(sudo:auth): User xguimard authenticated successfully via LLNG token
pam_llng:    pam_openbastion(sudo:account): Cached authorization for xguimard (TTL: 180 seconds)
```

Subsequent sudo calls over the next several minutes show **only the `account` phase — no `auth` phase, no OTP**:

```
12:16:15 sudo[199656]: pam_openbastion(sudo:account): Checking authorization for user: xguimard
12:18:36 sudo[199747]: pam_openbastion(sudo:account): Checking authorization for user: xguimard
12:19:13 sudo[199754]: pam_openbastion(sudo:account): Checking authorization for user: xguimard
12:19:55 sudo[199766]: pam_openbastion(sudo:account): Checking authorization for user: xguimard
```

## Root cause

This is **sudo's own credential cache** (`timestamp_timeout`, default 15 min, idle-based and **reset on each use** → effectively unlimited while the user keeps sudoing). When the timestamp is valid, sudo skips the PAM **auth** phase entirely, so `pam_openbastion` (and the OTP) is never invoked. open-bastion does **not** set `timestamp_timeout` anywhere (the generated `/etc/sudoers.d/open-bastion` only carries the `%open-bastion-sudo` rule), so the OS default applies.

## What is NOT the cause

- `pamAccessOfflineTtl` (LLNG) is unrelated: it only governs the **authorization** offline cache, used as a fallback **when the portal is unreachable**. With `pamAccessOfflineTtl=180` the bypass still happens, because the portal is up and the `account` phase re-verifies **live every sudo** (the `Cached authorization ... TTL` line is the store-after-live-success path, `!from_cache`).
- The **authorization** (\"sudo allowed\") is in fact re-checked live on every sudo, so an LLNG-side revocation of the right takes effect immediately. The gap is purely the **authentication** (OTP), which sudo caches.

## Proposed fix

Scope a `timestamp_timeout=0` to the SSO sudo group in the generated sudoers drop-in, so SSO users must present a fresh OTP on every sudo, while local break-glass admins keep normal sudo behavior:

```
Defaults:%open-bastion-sudo timestamp_timeout=0
```

To be added to `/etc/sudoers.d/open-bastion` by `ob-backend-setup`, `ob-bastion-setup` and the standalone setup (validated with `visudo -cf`, like the existing rule), plus the matching postinst/dry-run paths and docs.

## Notes

- This is a deliberate behavior change (OTP on every sudo for SSO users); worth calling out in the changelog/upgrade notes.
- Consider whether `sudo -i` / the `sudo-i` PAM service needs the same `Defaults:` scoping.

guimard:
Closes #178.

## The finding, restated

`sudo` keeps its own credential cache (`timestamp_timeout`, 15 min by default, idle-based and **rearmed on each use**). While it is valid, `sudo` skips the PAM `auth` phase entirely — `pam_openbastion` never runs, no LLNG one-time token is asked for. The issue's log excerpt shows exactly that: one `sudo:auth` line, then only `sudo:account` lines for minutes afterwards.

The issue's own analysis of what is *not* affected holds and is worth keeping in view: each token is single-use and consumed by the portal, and the `account` phase re-checks authorization live on every `sudo`, so an LLNG-side revocation takes effect immediately. What the cache weakens is the *freshness* claim, not authorization.

## The fix

`--enable-sudo-fresh-otp` on `ob-bastion-setup` and `ob-backend-setup` writes:

```
# /etc/sudoers.d/open-bastion
Defaults:%open-bastion-sudo timestamp_timeout=0
%open-bastion-sudo ALL=(ALL) ALL
```

Scoped to the SSO group, so local break-glass admins keep normal `sudo` behaviour — as the issue proposes.

**It is opt-in**, following the project's rule that invasive hardening is opted into rather than out of. Making it the default would start prompting on every `sudo` across an upgraded fleet, including inside scripts and long maintenance sessions, which is a behaviour change no operator asked for.

## Implementation notes

- The sudoers writer is split into `render_open_bastion_sudoers` (pure, testable without root) and `write_open_bastion_sudoers` (validates and installs), the same shape in both setup scripts.
- On a host that **already** has the drop-in, the flag only *adds* the `Defaults:` line and leaves the rest as the operator wrote it. Previously the whole block was skipped whenever the file existed, so an operator enabling this on an existing host would have got nothing.
- The result is still `visudo -cf` validated before installation, and the existing file is backed up first.
- `--dry-run` and the Mode E summary both report the state.

## Documentation

`doc/pam-modes.md` already described the trade-off (added in #227); it now names the supported way to close it, and **drops a wrong claim** that the setups regenerate the drop-in on every run — they only write it when absent, which is why the manual `open-bastion-local` workaround it suggested was never actually necessary.

EBIOS risk **R-S16** gains the same precision in French, with the flag stated as the condition for the literal Mode E claim.

## Tests

```
$ bash tests/test_ob_bastion_setup.sh   → 25/25   (+ fresh-OTP opt-in)
$ bash tests/test_ob_backend_setup.sh   → 21/21   (+ fresh-OTP opt-in)
$ bash tests/test_ob_pam_stacks.sh      → 28/28
```

The new case asserts both renderings, that the default carries no `timestamp_timeout` at all, that the flag is in `--help`, and — where `visudo` is available — that the opt-in drop-in parses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `auth required pam_deny.so`. Check with
   `grep '^auth' /etc/pam.d/sshd`.
 
+### Added
+
+- **`--enable-sudo-fresh-otp` on `ob-bastion-setup` and `ob-backend-setup`
+  (#178).** `sudo` keeps its own credential cache (`timestamp_timeout`, 15
+  minutes by default, idle-based and rearmed on each use). While it is valid,
+  `sudo` skips the PAM `auth` phase entirely, so `pam_openbastion` never runs and
+  no LLNG one-time token is asked for — a user who keeps elevating is not
+  prompted again, which is not what "a fresh SSO re-authentication for each
+  `sudo`" suggests.
+
+  What that does **not** weaken: each token is single-use and consumed by the
+  portal, and the `account` phase re-checks authorization live on every `sudo`,
+  so an LLNG-side revocation takes effect at once. What it weakens is the
+  freshness claim itself.
+
+  The new flag writes `Defaults:%open-bastion-sudo timestamp_timeout=0` into
+  `/etc/sudoers.d/open-bastion`, scoped to the SSO group so local break-glass
+  admins keep normal `sudo` behaviour. It stays **opt-in**: enabling it by
+  default would start prompting on every `sudo` across an upgraded fleet,
+  including inside scripts and long maintenance sessions. On a host that already
+  has the drop-in, the flag only *adds* the `Defaults:` line — the rest of the
+  file is left as the operator wrote it — and the result is `visudo -cf`
+  validated before installation, as before. `doc/pam-modes.md` and EBIOS risk
+  R-S16 now describe the trade-off and name the flag.
+
 ### Changed
 
 - **`ob-builder` artefacts that carry the client secret are no longer
--- a/doc/pam-modes.md
+++ b/doc/pam-modes.md
@@ -331,19 +331,27 @@ _token_ is useless (single use, short TTL), and a revoked _right_ is enforced
 immediately. What survives inside the window is the operator's own already
 authenticated terminal.
 
-If your policy requires a token prompt for **every** `sudo`, set
-`timestamp_timeout=0`. Put it in its own sudoers drop-in — `ob-bastion-setup`
-and `ob-backend-setup` regenerate `/etc/sudoers.d/open-bastion`, so anything
-written there is lost on the next run:
+If your policy requires a token prompt for **every** `sudo`, pass
+`--enable-sudo-fresh-otp` to `ob-bastion-setup` or `ob-backend-setup`. It scopes
+`timestamp_timeout=0` to the SSO group in `/etc/sudoers.d/open-bastion`, so SSO
+users go through the PAM `auth` phase — and therefore the LLNG token — on every
+elevation, while local break-glass admins keep normal `sudo` behaviour:
 
 ```bash
-cat > /etc/sudoers.d/open-bastion-local << 'EOF'
-# Require a fresh LLNG token for every sudo (no timestamp reuse).
-Defaults:%open-bastion-sudo timestamp_timeout=0
-EOF
-chmod 0440 /etc/sudoers.d/open-bastion-local
-visudo -c
+ob-bastion-setup --portal https://auth.example.com --max-security \
+                 --enable-sudo-fresh-otp
 ```
+
+```
+# /etc/sudoers.d/open-bastion
+Defaults:%open-bastion-sudo timestamp_timeout=0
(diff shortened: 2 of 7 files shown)


guimard:
Revue du 06/09/2026 — verdict : APPROVE (2 mineurs, 3 nits)

Opt-in conforme à la convention projet (`--enable-sudo-fresh-otp`, défaut inchangé), scopé au groupe SSO (admins break-glass épargnés), rendu testable sans root, validation `visudo -cf` + backup, fichier opérateur jamais réécrit. Offline fail-closed sans lockout : erreur serveur → refus propre sans incrément du rate-limiter, retry on-demand sur 401 (`src/pam_openbastion.c:3264-3285`). Pas de double prompt en Mode E. La doc promise par #178 est livrée (doc/pam-modes.md « sudo's timestamp cache », EBIOS R-S16, CHANGELOG) + correction d'une affirmation fausse pré-existante. Tests 25/25 + 21/21 + 28/28.

- [mineur] `scripts/ob-bastion-setup:1044` (idem backend:692) : l'opt-in est silencieusement annulable — sudoers applique « last match wins », et la ligne est *préfixée* : un `timestamp_timeout=15` existant plus bas (même fichier ou drop-in lu plus tard) écrase l'opt-in sans message. Fix : détecter un `timestamp_timeout` conflictuel pour le groupe et `warn`, ou apposer la ligne en fin de fichier.
- [mineur] `doc/pam-modes.md:356` : ne documente pas qu'avec le flag actif une panne du portail bloque 100 % des élévations sudo des utilisateurs SSO (vs 15 min de travail avec timestamp valide aujourd'hui). Une phrase suffit.
- [nit] Flag accepté mais inopérant sans bruit en deux configs (bastion sans `--max-security`, backend avec `--no-sudo`) → warn ; dry-run inexact sur hôte déjà configuré (« Would create … » vs ajout de la seule ligne Defaults) ; test `test_sudoers_rule_is_valid` ne capture plus la ligne `%open-bastion-sudo ALL=(ALL) ALL` (nom trompeur, couvert par ailleurs).

Issue #244, opened 2026-09-06
Flaky test_offline_cache in the RPM job, and we cannot diagnose it: re-running the job destroys the only log


guimard:
`test_offline_cache` failed once in the `Build RPM (Rocky Linux 9)` job of CI, then passed on a re-run of the very same commit with no code change. It is a flake, and — worse — one we currently cannot diagnose after the fact.

## What was observed

PR #237, run [34019793647](https://github.com/linagora/open-bastion/actions/runs/34019793647), job `Build RPM (Rocky Linux 9)` (attempt 1):

```
13/21 Test #13: test_offline_cache ...............***Failed    5.03 sec
...
95% tests passed, 1 tests failed out of 21
The following tests FAILED:
	 13 - test_offline_cache (Failed)
error: Bad exit status from /var/tmp/rpm-tmp.F4O3ls (%check)
```

The other twelve checks on the same commit were green, and `Build RPM (Rocky Linux 9)` was green on `main` at the same base commit. `gh run rerun --failed` on the identical tree gave `13/21 Test #13: test_offline_cache ... Passed 7.30 sec` and the job succeeded.

## The evidence is gone

`rpm/open-bastion.spec` runs `ctest --output-on-failure --verbose`, so the failing sub-test's name and its `Assertion failed: ... at line N` line **were** in the attempt-1 log. Re-running the failed job replaced that log: `GET /actions/jobs/101450328558/logs` and `.../runs/34019793647/attempts/1/jobs` now only serve the successful attempt. We therefore do not know which of the 17 sub-tests failed.

That is the first thing to fix — see "Proposed work" below.

## What has been ruled out

Investigated locally, none of these explain it:

[...]

PR #245, opened 2026-09-06
ci: keep the ctest log when a test fails, and make the concurrency test explain itself

guimard:
Follow-up to #244, which this PR does **not** close: the flake itself is still unexplained. What this changes is that the next occurrence will be diagnosable.

## Why

`test_offline_cache` failed once in `Build RPM (Rocky Linux 9)` on #237 ([run 34019793647](https://github.com/linagora/open-bastion/actions/runs/34019793647)) and passed on a re-run of the identical commit. `rpm/open-bastion.spec` runs `ctest --output-on-failure --verbose`, so the failing sub-test's name and its `Assertion failed: ... at line N` **were** in the attempt-1 log — but GitHub serves only the latest attempt of a job, so re-running to confirm the flake is what destroyed the evidence. There is no way back to it.

## What it does

**Keep the log.** The three jobs that run `ctest` — `Build and Test` (both matrix legs), `Build Debug with Sanitizers`, `Build RPM` — upload `Testing/Temporary/` as an artifact on step failure. `LastTest.log` holds the test binary's own stdout; `LastTestsFailed.log` holds which tests failed. For the RPM job the ctest run happens inside the rpmbuild tree, so the path is `~/rpmbuild/BUILD/open-bastion-*/redhat-linux-build/Testing/Temporary/` — taken from the real build tree in the failing log, not guessed.

**Make the concurrency test say why it failed.** `test_concurrent_failed_attempts` (the `#186` lockout regression) had four failure paths that printed nothing and returned a bare `FAILED`:

- `pipe()` fails
- `fork()` fails — plausible under container memory pressure, and it silently left `started < 6`
- a short `write()` to the release barrier
- `offline_cache_get_entry()` fails

Each now names what went wrong. In addition each child reports through its exit status whether its `verify()` actually returned `OFFLINE_CACHE_ERR_PASSWORD`, so a counter shortfall is attributed to a specific cause instead of surfacing as an unexplained `failed_attempts=5 expected=6`. Abnormal child termination (signal) is reported too.

## Verification

```
$ ./tests/test_offline_cache
Results: 17/17 tests passed
```

Fault injection — children sending the *correct* password instead of the wrong one:

```
Testing concurrent_failed_attempts... (a child's verify did not report a wrong password) x6 (failed_attempts=0 expected=6) FAILED
```

`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` passes.

## What this does not do

It does not fix the flake. It did not reproduce in 40 local runs of the suite, nor in 8 runs with the machine oversubscribed 3× (24 spinners on 8 cores). #244 records what was ruled out — notably the 1-second TTL boundary race, which measurement excludes: `store()` takes its `now` *after* the Argon2id hash and `verify()` takes its own *before* its hash, so the window between the two `time()` calls is a few milliseconds, not the ~270 ms Argon2id costs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,11 @@ jobs:
       contents: read
 
     strategy:
+      # A ctest failure on one leg must not cancel the other: a cancelled step
+      # never runs `if: failure()`, so the sibling's LastTest.log would be
+      # dropped — and for a timing-sensitive flake (#244) a second sample from
+      # a different runner is the most useful evidence there is.
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04, ubuntu-24.04]
 
@@ -39,7 +44,7 @@ jobs:
           jq
 
     - name: Configure CMake
-      run: cmake -B build -DCMAKE_BUILD_TYPE=Release -DUSE_LIBSODIUM=ON
+      run: cmake -B build -DCMAKE_BUILD_TYPE=Release -DUSE_LIBSODIUM=ON -DINSTALL_DESKTOP=ON
 
     - name: Build
       run: cmake --build build --parallel
@@ -53,10 +58,10 @@ jobs:
     # re-running a failed job replaces the workflow log, and with it the only
     # record of which sub-test failed and on which assertion (#244).
     - name: Upload ctest logs on failure
-      if: failure()
+      if: failure() || cancelled()
       uses: actions/upload-artifact@v7
       with:
-        name: ctest-logs-${{ matrix.os }}
+        name: ctest-logs-${{ matrix.os }}-${{ github.run_attempt }}
         path: build/Testing/Temporary/
         retention-days: 7
         if-no-files-found: warn
@@ -107,6 +112,7 @@ jobs:
         cmake -B build \
           -DCMAKE_BUILD_TYPE=Debug \
           -DUSE_LIBSODIUM=ON \
+          -DINSTALL_DESKTOP=ON \
           -DCMAKE_C_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer"
 
     - name: Build
@@ -120,10 +126,10 @@ jobs:
         ctest --output-on-failure
 
     - name: Upload ctest logs on failure
-      if: failure()
+      if: failure() || cancelled()
       uses: actions/upload-artifact@v7
       with:
-        name: ctest-logs-sanitizers
+        name: ctest-logs-sanitizers-${{ github.run_attempt }}
         path: build/Testing/Temporary/
         retention-days: 7
         if-no-files-found: warn
@@ -176,6 +182,8 @@ jobs:
     container: rockylinux:${{ matrix.version }}
 
     strategy:
+      # Same reason as the build matrix: keep every leg's ctest log (#244).
+      fail-fast: false
       matrix:
         version: [9]
 
@@ -211,10 +219,10 @@ jobs:
     # The %check section runs ctest inside the rpmbuild tree, so its
     # LastTest.log lives there and is lost with the job log on a re-run (#244).
     - name: Upload ctest logs on failure
-      if: failure()
+      if: failure() || cancelled()
       uses: actions/upload-artifact@v7
       with:
-        name: ctest-logs-rocky${{ matrix.version }}
+        name: ctest-logs-rocky${{ matrix.version }}-${{ github.run_attempt }}
         path: ~/rpmbuild/BUILD/open-bastion-*/redhat-linux-build/Testing/Temporary/
         retention-days: 7
         if-no-files-found: warn
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,18 @@ jobs:
         cd build
         ctest --output-on-failure
 
+    # ctest writes the failing test's own stdout to LastTest.log. Keep it:
+    # re-running a failed job replaces the workflow log, and with it the only
+    # record of which sub-test failed and on which assertion (#244).
+    - name: Upload ctest logs on failure
+      if: failure()
+      uses: actions/upload-artifact@v7
+      with:
+        name: ctest-logs-${{ matrix.os }}
+        path: build/Testing/Temporary/
+        retention-days: 7
+        if-no-files-found: warn
+
     - name: Check PAM module symbols
       run: |
         echo "Checking exported PAM symbols..."
@@ -106,6 +118,15 @@ jobs:
       run: |
         cd build
         ctest --output-on-failure
+
+    - name: Upload ctest logs on failure
+      if: failure()
+      uses: actions/upload-artifact@v7
+      with:
+        name: ctest-logs-sanitizers
+        path: build/Testing/Temporary/
+        retention-days: 7
+        if-no-files-found: warn
 
   package-deb:
     name: Build Debian Package
@@ -186,6 +207,17 @@ jobs:
           --transform "s,^,open-bastion-${VERSION}/," \
           --exclude='.git' .
         rpmbuild -ba rpm/open-bastion.spec
+
+    # The %check section runs ctest inside the rpmbuild tree, so its
+    # LastTest.log lives there and is lost with the job log on a re-run (#244).
+    - name: Upload ctest logs on failure
+      if: failure()
+      uses: actions/upload-artifact@v7
+      with:
+        name: ctest-logs-rocky${{ matrix.version }}
+        path: ~/rpmbuild/BUILD/open-bastion-*/redhat-linux-build/Testing/Temporary/
+        retention-days: 7
+        if-no-files-found: warn
 
     - name: Upload RPM package
       uses: actions/upload-artifact@v7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,13 +132,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ctest` now upload `Testing/Temporary/` as an artifact when the test step
   fails, so the evidence survives a re-run.
 
+  Two of the three jobs could not have held that evidence: `test_offline_cache`
+  is only compiled with `INSTALL_DESKTOP=ON` (`tests/CMakeLists.txt:123`), which
+  neither the build matrix nor the sanitizer job passed. Both now do, so the
+  test of #244 also runs under ASan/UBSan — where a concurrency bug is most
+  likely to be caught — and on two more distributions. The artifact names carry
+  `github.run_attempt`, because artifacts are immutable per run and a partial
+  re-run keeps attempt 1's: a fixed name would make the re-run's upload fail
+  with `409 Conflict` and lose exactly the confirmation evidence #244 is about.
+  `fail-fast: false` on both matrices, because a cancelled step never runs
+  `if: failure()`, so one leg's failure would silently drop its sibling's log.
+
   `test_concurrent_failed_attempts` (the `#186` lockout regression) also had
-  four failure paths that printed nothing: a failed `pipe()` or `fork()`, a
-  short write to the release barrier, and a failed `offline_cache_get_entry()`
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -49,6 +49,18 @@ jobs:
         cd build
         ctest --output-on-failure
 
+    # ctest writes the failing test's own stdout to LastTest.log. Keep it:
+    # re-running a failed job replaces the workflow log, and with it the only
+    # record of which sub-test failed and on which assertion (#244).
+    - name: Upload ctest logs on failure
```
guimard, line 60: This step fails in the PR's own scenario. With a fixed `name` and `overwrite: false` (the v7 default), attempt 1's failure uploads the artifact; if a "Re-run failed jobs" (partial re-run) fails again, upload-artifact gets `409 Conflict: an artifact with this name already exists on the workflow run` — artifacts are immutable per run, and partial re-runs keep attempt-1's artifacts. The attempt-2 log (the confirmation evidence for #244) is never stored, and every later attempt keeps 409ing. Add `overwrite: true`, or make the name attempt-unique: `ctest-logs-${{ matrix.os }}-${{ github.run_attempt }}`. Same fix needed in the sanitizer and Rocky steps.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,18 @@ jobs:
...
+      if: failure()
```
guimard: `fail-fast` defaults to `true` and isn't set here, so a ctest failure on one leg cancels the other leg mid-run, and a cancelled step's outcome is `cancelled`, not `failure` — `if: failure()` never runs on that leg, and its `LastTest.log`/`LastTestsFailed.log` is dropped. For a timing-sensitive flake like #244, the sibling leg's sample is exactly the multi-sample evidence this PR wants to keep. Suggest `fail-fast: false`, and optionally `if: failure() || cancelled()` — ctest updates `LastTestsFailed.log` incrementally, so archiving on cancellation preserves whatever the killed run had already observed.

tests/test_offline_cache.c
```diff
@@ -587,34 +594,61 @@ static int test_concurrent_failed_attempts(void)
             } while (r < 0 && errno == EINTR);
             close(barrier[0]);
 
-            offline_cache_verify(cache, "concurrent_user", "wrong_pass", NULL);
-            _exit(0);
+            /* The exit status carries the verdict: a child whose verify did
```
guimard: This verdict channel can't fire in the mode this comment targets. In the #186 lost-increment mode — `lock_cache_entry()` fails and the read-modify-write runs unlocked (the documented degraded behaviour, src/offline_cache.c:399-406), or the final `update_cache_entry()` write fails with its return value discarded (src/offline_cache.c:1269) — every child's verify still returns `OFFLINE_CACHE_ERR_PASSWORD` (the wrong-password branch does so unconditionally), so all children exit 0 and the output stays the bare `(failed_attempts=5 expected=6)`. The channel only fires when a child returns a non-PASSWORD error, which the exact-N counter assertion already catches. It does add real value for killed/crashed children (`WIFSIGNALED`). To make it genuinely diagnostic, exit with a cret-derived code (`_exit(cret)`) and decode via `offline_cache_strerror()` in the parent — or soften this comment and the CHANGELOG claim.

tests/test_offline_cache.c
```diff
@@ -587,34 +594,61 @@ static int test_concurrent_failed_attempts(void)
...
 
-    int ok = (started == OC_CONC_CHILDREN);
+    if (started != OC_CONC_CHILDREN) {
+        printf("(started=%d expected=%d) ", started, OC_CONC_CHILDREN);
+        ok = 0;
+    }
 
     offline_cache_entry_t entry;
-    if (ok && offline_cache_get_entry(cache, "concurrent_user", &entry) == OFFLINE_CACHE_OK) {
+    int gret = offline_cache_get_entry(cache, "concurrent_user", &entry);
```
guimard, line 662: When `started < OC_CONC_CHILDREN` (fork failure), this check compares `failed_attempts` against `OC_CONC_CHILDREN` and prints a misleading second layer: the `started` children that ran each incremented correctly, so `(failed_attempts=3 expected=6)` falsely reads as a cache-integrity regression on top of the already-reported fork failure. Compare against `started` instead — identical semantics when `started == OC_CONC_CHILDREN`, and the correct baseline otherwise.

CHANGELOG.md
```diff
@@ -124,6 +124,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  failed (#244).** `test_offline_cache` failed once in the Rocky 9 RPM job and
+  passed on a re-run of the same commit — but the re-run replaced the workflow
+  log, which was the only record of which sub-test failed and on which
+  assertion, so the flake could not be diagnosed at all. The three jobs that run
+  `ctest` now upload `Testing/Temporary/` as an artifact when the test step
+  fails, so the evidence survives a re-run.
+
+  `test_concurrent_failed_attempts` (the `#186` lockout regression) also had
+  four failure paths that printed nothing: a failed `pipe()` or `fork()`, a
+  short write to the release barrier, and a failed `offline_cache_get_entry()`
```
guimard: The barrier-write bullet misdescribes the old code: on a failed/short barrier write the old code broke out of the release loop and `close(barrier[1])` released the remaining children via EOF — they still ran verify(), all six increments landed, `started == 6`, and the old test PASSED silently. That path was a silent false-PASS, not a "returned FAILED with no reason" (only pipe()/fork()/get_entry did). Worth correcting since this changelog is the #244 archaeology record — and note the new code's `ok = 0` on that path is a verdict change, not just added noise.

.github/workflows/ci.yml
```diff
@@ -107,6 +119,15 @@ jobs:
         cd build
         ctest --output-on-failure
 
+    - name: Upload ctest logs on failure
```
guimard, line 128: Coverage note: `test_offline_cache` (and thus `test_concurrent_failed_attempts`) is only compiled with `INSTALL_DESKTOP=ON` (tests/CMakeLists.txt:123), which only rpm/open-bastion.spec passes — this job and the matrix job configure without it. So these two upload steps can never hold the log of the very test this PR targets; only the Rocky step can. And the new C diagnostics will never run under ASan/UBSan, where the concurrency behaviour is most likely to misbehave. Worth either enabling `-DINSTALL_DESKTOP=ON` here, or at least saying in the CHANGELOG that only the Rocky job protects the #244 evidence.

guimard (commented):
Reviewed (multi-angle + verification pass). The log-keeping idea is right, but two CI details break the PR's own scenarios and the C diagnostics claim is overstated.

**Must-fix**

1. **Fixed artifact names + `overwrite: false` → 409 on re-run.** Attempt 1 fails and uploads the artifact; a partial re-run ("Re-run failed jobs" — the exact #244 flow) that fails again gets `409 Conflict: an artifact with this name already exists on the workflow run`, because artifacts are immutable per run and partial re-runs keep attempt-1's artifacts. The attempt-2 evidence is never stored. Fix: `overwrite: true`, or `-${{ github.run_attempt }}` in the name (same fix as apache/datafusion-comet#3723 and nasa/fprime-actions#11).
2. **`fail-fast` cancels the sibling matrix leg** and a cancelled step never runs `if: failure()` (outcome is `cancelled`, not `failure`) — the second leg's log is dropped, which is exactly the multi-sample evidence a timing-sensitive flake needs. Fix: `fail-fast: false`, optionally `failure() || cancelled()` to also archive partial logs of a mid-ctest cancellation (`LastTestsFailed.log` is updated incrementally).

**Should-fix**

3. **Exit-status verdict overstated** (inline comment on the child block): in the #186 lost-increment mode every child still exits 0, so the bare `(failed_attempts=5 expected=6)` the PR set out to replace is unchanged there. Suggest exiting with a cret-derived code, or softening the comment/CHANGELOG.
4. **`failed_attempts` expected value** (inline): compare against `started`, not `OC_CONC_CHILDREN` — avoids a misleading `(failed_attempts=3 expected=6)` after a fork failure.
5. **CHANGELOG barrier-write claim** (inline): that path previously produced a silent false-PASS, not a bare FAILED.

**Notes (no action needed unless you want to)**

- Coverage gap (inline): `test_offline_cache` is only built with `INSTALL_DESKTOP=ON`, which only the RPM job passes — the two host-job upload steps can never hold the #244 test's log.
- Killed-child → new failure mode is defensible (a signal-killed child is worth failing on, esp. under ASan); the added surface is tiny.
- `wait()` should retry on EINTR before consuming a slot (unreachable today — no signal handlers — but a one-line future-proofing).
- Rocky `path:` is coupled to the spec's `%check` layout; `if-no-files-found: warn` would hide a future break. Acceptable as-is; a gated assertion step would harden it.
- The three near-identical upload blocks: a composite action would be disproportionate; leave as-is.

I'd ask for changes on items 1–2 (they break the stated purpose), and 3–5 are small fixes or wording.

guimard:
Les six points tiennent, vérification faite. Corrigés en `a3092c8`.

**1. Noms d'artefacts (must-fix)** — confirmé, et j'ai pris `-${{ github.run_attempt }}` plutôt que `overwrite: true`. `overwrite` réglerait le 409 en **supprimant le log de la tentative 1**, ce qui est exactement l'inverse du propos de #244 : c'est la première tentative qui porte la preuve du flake, la seconde ne fait que confirmer. Le suffixe garde les deux.

**2. `fail-fast` (must-fix)** — confirmé, aucune des deux matrices ne le posait (l.20 et l.178). `fail-fast: false` sur les deux, et `if: failure() || cancelled()` sur les trois étapes d'upload comme suggéré.

**3. Verdict par code de sortie** — tu as raison et je m'étais trompé. `offline_cache.c:1268` jette le retour de `update_cache_entry()` puis retourne `ERR_PASSWORD` inconditionnellement : dans le mode perte-d'incrément, tous les enfants sortent 0 et l'affichage reste le `(failed_attempts=5 expected=6)` nu. Le commentaire et le CHANGELOG disent maintenant ce que le canal couvre — erreur précoce, enfant tué — et rien de plus. J'ai suivi ta suggestion de dériver le code de `cret`, avec un décodage par `offline_cache_strerror()` côté parent et 127 réservé au seul cas qui ne doit jamais passer pour un succès : un mauvais mot de passe accepté.

Mutation-test : en pointant un enfant sur un utilisateur inexistant, la sortie passe de `(failed_attempts=5 expected=6) FAILED` à `(a child's verify returned Entry not found) (failed_attempts=5 expected=6) FAILED`.

**4. Comparaison à `started`** — corrigé, avec le commentaire qui dit pourquoi.

**5. Le barrier était un faux PASS, pas un FAILED muet** — vérifié sur `main` : le `break` sortait de la boucle, `close(barrier[1])` libérait les enfants restants par EOF, les six incréments atterrissaient, `started == 6`, test **PASSED**. Mon texte était faux, et sur un changelog qui sert d'archéologie de #244 ça compte. Réécrit, et j'ai noté explicitement que le `ok = 0` sur ce chemin est un **changement de verdict**, pas du bruit en plus.

**6. Trou de couverture** — confirmé, et c'était le plus gênant : `INSTALL_DESKTOP` n'est passé que par `debian/rules` et `rpm/open-bastion.spec`, et le job Debian fait `cpack`, pas `debian/rules`, donc **seul le job Rocky** compilait le test. Deux de mes trois étapes d'upload ne pouvaient rien contenir. J'ai choisi la version haute plutôt que d'amender le CHANGELOG : `-DINSTALL_DESKTOP=ON` sur la matrice et sur le job sanitizers, ce qui fait enfin tourner le test de #244 sous ASan/UBSan.

Les trois autres notes : `wait()`/EINTR et l'assertion sur le chemin Rocky, je les laisse — d'accord sur le fond, mais hors du propos de cette PR ; l'action composite, d'accord, disproportionnée.

Le flake lui-même n'est toujours pas corrigé, et c'est délibéré : l'hypothèse de course sur la frontière TTL est réfutée par la mesure (fenêtre réelle de quelques ms), consignée comme écartée dans #244.

PR #264, opened 2026-09-07
fix(service-accounts): ship ob-service-account-keys, and stop the docs denying it is needed

guimard:
Batch 1 of #263 — the part that needs no design decision.

## Root cause

`scripts/ob-service-account-keys` is the `AuthorizedKeysCommand` helper that makes a
service account usable. It is in the tree and in **none** of the three packaging
paths: absent from `CMakeLists.txt`, from `debian/open-bastion.install` and from
`rpm/open-bastion.spec`. A `.deb` built from `main` carries no copy.

So there was nothing on a host to point `AuthorizedKeysCommand` at, and everyone who
needed it installed their own, at a path of their choosing:

| where | path |
| --- | --- |
| `docker-demo-maxsec`, `docker-demo-token-svc` | `/usr/local/bin/` |
| `local-test/`, `doc/service-accounts.md` | `/usr/local/sbin/` |
| the package | *nowhere* |

The demos work because they do by hand exactly what the reporter had to do from
`/tmp`. That is why nothing ever caught it.

It now installs to `/usr/sbin/ob-service-account-keys` — a package must not write
under `/usr/local` (Debian policy 9.1.2) — and every shipped reference converges
there. The two demos drop their hand-install; `make install` provides it.

## The documentation was worse than incomplete

Both passages are the ones an operator reads *before* deploying.

**`doc/service-accounts.md`, Mode E:** *"No `authorized_keys` file is required."*
Literally true, and it reads as "nothing else is needed". Without an
`AuthorizedKeysCommand`, sshd has no key for the account and refuses at the protocol
layer — `pam_sm_authenticate` is never reached. The paragraph presented the PAM
fingerprint check as the authorisation. It is a **re-validation**: it can refuse a key
sshd accepted, but it cannot make sshd accept a key it has no record of. This is the
reasoning that cost the reporter a debugging session.

**And:** *"`ExposeAuthInfo yes` must be set (the setups do this)."* They write it only
inside `configure_max_security_sshd()` (`ob-bastion-setup:764`, `ob-backend-setup:631`),
which returns immediately without `--max-security`. On a bastion that is not in Mode E
— the reporter's case — it is never written.

## Evidence

`tests/test_ob_service_account_keys.sh` pins the packaging, the single path, both
documentation claims, and — deliberately — that `ExposeAuthInfo` **is** still Mode-E
only, because the corrected paragraph rests on that and would become the wrong one if a
setup script started writing it unconditionally.

[...]

Issue #263, opened 2026-09-07
ob-builder: generated service accounts are not usable without additional host-side SSH configuration


kvn-chn:
**Summary**

When defining a service account through ob-builder, the generated deployment artifact creates /etc/open-bastion/service-accounts.conf, but the account is not immediately usable for SSH authentication.
Additional manual configuration is required on the target host:
- Install the AuthorizedKeysCommand helper.
- Install the service account public key.
- Configure sshd to use the helper.
- Enable ExposeAuthInfo yes.
- Ensure the Unix account and its UID/GID are correctly resolvable.
This was not obvious from the generated deployment flow and initially resulted in a service account that could not connect.

**Environment**

- Open Bastion version/commit: v0.6.2
- Deployment method: ob-builder generated artifact
- Target distribution: ansible-lead-adm vm
- Target role: bastion
- Service account: ansible

**Configuration**

The service account was configured with a policy similar to:

```
[ansible]
key_fingerprint = SHA256:...
shell = /bin/bash
home = /home/ansible
uid = 6001
gid = 6001
```
The target already contained an ansible Unix account:
`ansible:x:6001:6001:ansible-bastion:/home/ansible:/bin/bash`
However, the group named ansible had GID 6100, while the account’s primary GID was 6001. We created a group for GID 6001 and assigned it as the account’s primary group:
```
groupadd --gid 6001 ansible-primary
usermod --gid ansible-primary ansible
```

**Observed Behavior**

The service account could not authenticate using its SSH key after deploying the generated artifact.
The generated artifact configured the Open Bastion service-account policy, but did not configure the SSH layer to accept the public key. In particular, the following were missing:
```
/usr/local/sbin/ob-service-account-keys
/etc/open-bastion/service-accounts.d/ansible.pub
/etc/ssh/sshd_config.d/09-open-bastion-service-keys.conf
```
Without this configuration, sshd rejects the key before Open Bastion PAM validation can succeed.
A second issue can occur when the account does not already exist locally: nss_openbastion only resolves service accounts before authentication when both uid and gid are defined. Otherwise, sshd can reject the user as unknown before PAM has a chance to create it.

**Workaround**

We manually completed the SSH-layer configuration on the remote host.
Install the helper:
```
install -o root -g root -m 0755 \
  /tmp/ob-service-account-keys \
  /usr/local/sbin/ob-service-account-keys
```
Install the public key:
```
install -d -o root -g root -m 0755 \
[...]

Issue #257, opened 2026-09-07
ob-session-monitor terminates live sessions when it merely fails to reach /pam/userinfo


guimard:
## The problem

`check_user_valid()` in `scripts/ob-session-monitor` ends its portal call with:

```sh
    response=$(curl -sf -m 10 \
        ... "${PORTAL_URL}/pam/userinfo" 2>/dev/null) || return 1
```

`curl -sf` returns non-zero for a connection failure, a DNS failure, a TLS
failure, the 10-second timeout **and any HTTP status >= 400**. All of them
become `return 1`.

The caller reads that as an authoritative verdict (`:262`):

```sh
        else
            # User no longer valid - terminate session
            log_crit "User $user no longer valid on LLNG, terminating session"
            ... loginctl terminate-session "$session_id" ...
            rm -f "$MARKER_DIR/$user"
```

So a transient failure on one request **terminates that user's live sessions**
and deletes their offline marker, while logging the false statement "no longer
valid on LLNG".

## Why the existing reachability gate does not cover this

`revalidate_sessions` only runs when the portal probe succeeded (`:376`), so
this is not reachable during a plain outage. That is worth stating plainly
because it narrows the window — but it does not close it, for two reasons:

1. **The probe checks a different endpoint.** `check_portal()` (`:113`) fetches
   `${PORTAL_URL}/desktop/login?check=1`. A healthy portal front end says
   nothing about `/pam/userinfo`: the `pam-access` plugin erroring, a 500/502/503
   from that route, a per-endpoint rate limit, or a `SERVER_TOKEN` that has
   expired — a 401 — all pass the probe and fail the call.
2. **It bypasses the deliberate grace period.** `handle_sso_unreachable()`
   (Part D, `:283`) exists precisely so that losing the portal terminates
   sessions only after `MAX_SSO_UNREACHABLE`, one hour by default. Terminating
   on a single failed request skips that hour entirely — the design decided how
   long to tolerate an unreachable portal, and this path does not honour it.

The expired-`SERVER_TOKEN` case is not hypothetical: the heartbeat timer failing
to arm, and the server token dying overnight as a result, is a bug this project
has already shipped and fixed once (#159). The same failure now reaches a code
path that terminates sessions.

## What the fix has to distinguish

"The portal says this user is gone" and "we could not ask the portal" are
different answers and must not share a return code. The shape used for the
signing failure a few lines above (`:156`-`:162`, added in #247) is the
precedent:

```sh
[...]

PR #265, opened 2026-09-07
feat(setup): --enable-service-keys, so a service account can actually log in

guimard:
Batch 2 of #263.

## Why

`service-accounts.conf` tells `pam_openbastion` which accounts exist and what
fingerprint each should present. It never made the key usable. sshd needs an
authorized key, and on a bastion there is no `~/.ssh/authorized_keys` to read —
Mode E sets `AuthorizedKeysFile none`, and the account often does not exist until
first login. Without an `AuthorizedKeysCommand`, sshd refuses at the protocol layer
and PAM never runs. That is how the field report arrived.

`--enable-service-keys` (both setup scripts) writes the whole drop-in and creates
`service-accounts.d` 0755 so the `AuthorizedKeysCommandUser` can traverse it.

## `ExposeAuthInfo` is part of the flag, on purpose

Leaving it to Mode E would have shipped the same defect one layer up.

It is what puts `SSH_USER_AUTH` in the PAM environment, and without it the
fingerprint check has nothing to read for a plain public key — the principals spool
cannot substitute, since sshd runs `AuthorizedPrincipalsCommand` only for
certificate sessions. An `AuthorizedKeysCommand` alone would mean **sshd accepts the
key while `key_fingerprint` is never verified**: exactly the silent
half-configuration `doc/service-accounts.md` warns about, installed by us this time.

Conversely, Mode E writes `ExposeAuthInfo` but *not* the `AuthorizedKeysCommand`, so
`--max-security` alone still leaves a service account unable to log in. The
documentation now says both.

Opt-in, per this project's rule for changes to system-wide behaviour: it alters sshd
for every session on a host that may have no service account at all.
`fingerprint_required` stays a separate decision — it applies to every SSH login,
not only service accounts — so the flag **reports** whether it is set rather than
setting it.

## A test caught me invalidating my own documentation

`doc/service-accounts.md` said `ExposeAuthInfo` is written "only in Mode E". Adding a
second writer makes that false, and the pin I added in #264 failed on it —
correctly. The assertion now covers the property that survives: **every**
`ExposeAuthInfo` write sits behind an opt-in guard, none is unconditional. Writing it
unconditionally in `configure_sshd` is caught.

## Evidence

The real function driven in a `debian:trixie` container, both ways:

```
### OFF:  RC=0        drop-in: absent
### ON:   RC=0        WARN: fingerprint_required is not set …
   AuthorizedKeysCommand /usr/sbin/ob-service-account-keys %u
   AuthorizedKeysCommandUser nobody
   ExposeAuthInfo yes
   keys dir: root 755
```

Then with `openssh-server` installed — because a malformed drop-in would break sshd
outright, not just this feature:

```
sshd ACCEPTS the configuration
exposeauthinfo yes
authorizedkeyscommand /usr/sbin/ob-service-account-keys %u
authorizedkeyscommanduser nobody
```

- 35 shell suites, **none skipped**; shellcheck unchanged from `main`; prettier clean.

## What is left of #263

Batch 3: `ob-builder` learning a `public_key` field so it can deploy `<name>.pub`
itself. Its record schema carries only the fingerprint today — and once the key is
there, the fingerprint can be derived from it rather than transcribed by hand, which
removes a class of copy errors.


scripts/ob-bastion-setup
```diff
@@ -2169,6 +2274,7 @@ main() {
     # From this point on, a failure may require restoring files from
     # $BACKUP_DIR by hand.
     configure_sshd || exit 1
+    configure_service_account_keys || exit 1
```
guimard, line 2410: `configure_sshd` runs `sshd -t` at :404 (returning 1 on failure) and runs BEFORE this call. On an upgrade host carrying a hand-assembled `09-open-bastion-service-keys.conf` with a syntax error — the exact population this flag targets, since the pre-PR doc told operators to write that file by hand — the run dies at `configure_sshd || exit 1` on "SSH configuration is invalid!" before this function ever overwrites the broken file. `restart_sshd` is skipped, and the next reboot applies the broken file, killing sshd. The repair path the flag adds is unreachable exactly when the file it would fix is what breaks validation. Consider validating after the rewrite instead (or letting the writer run before the gate). Same ordering in `ob-backend-setup` (:2264-2265).

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+
+    backup_file "$drop_in"
+    cat > "$drop_in" <<SVCKEYS
+# Open Bastion service-account SSH layer
+# Generated by $PROG_NAME on $(date)
+#
+# Serves /etc/open-bastion/service-accounts.d/<user>.pub to sshd, so a service
+# account can authenticate with its key without a ~/.ssh/authorized_keys and
+# without the Unix account existing yet. See doc/service-accounts.md.
+AuthorizedKeysCommand $helper %u
```
guimard, line 919: This drop-in silently overrides any pre-existing `AuthorizedKeysCommand` in the main sshd_config: `configure_sshd` inserts `Include /etc/ssh/sshd_config.d/*.conf` at line 1 (:372) and sshd keeps the FIRST value per keyword — the code's own comment at :377-380 states that rule. A host already running its own key server (`AuthorizedKeysCommand /usr/local/bin/ldap-keys %u` in sshd_config) — precisely the #263 audience — gets that line made dead after a flagged run: every previously key-authorized user silently loses access. Nothing probes for an existing command, and no `sshd -t`/`-T` runs after this write. At minimum, warn when one is detected. Same in `ob-backend-setup` (:694).

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+        return 0
+    fi
+
+    if [ ! -x "$helper" ]; then
+        error "$helper not found. It ships with the open-bastion package;"
+        error "reinstall it, or drop --enable-service-keys."
+        return 1
+    fi
+
+    if [ ! -d "$SSHD_CONFIG_DIR" ]; then
```
guimard, line 875: These two `return 1` paths become `configure_service_account_keys || exit 1` in main() (:2277): a hard abort mid-Phase-3, AFTER `configure_sshd` has already rewritten sshd_config to cert-only (`PasswordAuthentication no`) and enrollment completed, but BEFORE `configure_pam_sshd` and `restart_sshd`. Both sibling writers degrade gracefully on the same condition — `configure_sshd` and `configure_max_security_sshd` fall back to appending to `$SSHD_CONFIG` when the dir is absent — so the toolchain demonstrably supports drop-in-less hosts; this function instead leaves one half-locked (the next restart applies cert-only auth without the PAM authorization layer installed). The two preconditions are cheap static checks that could run before the point of no return. Same in `ob-backend-setup` (:675-679).

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+# ExposeAuthInfo is written HERE and not left to Mode E. It is what puts
+# SSH_USER_AUTH in the PAM environment, and without it the fingerprint check
+# pam_openbastion is supposed to do for a plain public-key login has nothing to
+# read -- the principals spool cannot fill the gap either, since sshd runs
+# AuthorizedPrincipalsCommand only for certificate sessions. Shipping the
+# AuthorizedKeysCommand without ExposeAuthInfo would mean sshd accepts the key
+# and key_fingerprint is never enforced: the silent half-configuration that
+# doc/service-accounts.md now warns about.
+configure_service_account_keys() {
+    if [ "$ENABLE_SERVICE_KEYS" != "true" ]; then
```
guimard, line 854: The flag is one-way. When it is off this is a bare early return, and nothing anywhere removes a previously written `09-open-bastion-service-keys.conf` (`git grep 09-open-bastion` finds only the two writers plus local-test). Re-running setup WITHOUT the flag — this project's documented "re-run setup = fix" path — leaves `AuthorizedKeysCommand` + `ExposeAuthInfo yes` live for every session; converting a host to Mode E with `--max-security` keeps service-account pubkey logins working at the protocol layer, contradicting the CHANGELOG promise that "Mode E alone still leaves a service account unable to log in" (`AuthorizedKeysFile none` does not disable AuthorizedKeysCommand). The suite sets the bar itself: "without the flag, nothing must change" (tests:311-313). The `rm -f` precedent for the legacy 50- drop-in exists at :384. Consider removing the generated file when the flag is off. Same in `ob-backend-setup` (:654).

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+        return 1
+    fi
+
+    # 0755: the AuthorizedKeysCommandUser (nobody) has to traverse and read it.
+    # The package creates it; assert it here because an older host upgraded
+    # from a release that did not ship it will not have it.
+    install -d -m 0755 -o root -g root "$keys_dir"
+
+    backup_file "$drop_in"
+    cat > "$drop_in" <<SVCKEYS
```
guimard: This write chain is unguarded, and it cannot fail loudly: main() calls this function as `configure_service_account_keys || exit 1`, and bash disables errexit for the entire body of a function executed in an `||` list — only the explicit `return 1` guards abort. A failing `cat` (full disk, read-only /etc/ssh/sshd_config.d) falls through to `track_created` + `info "Created $drop_in"`, and the function's last command is the `if ! grep … fi` tail (:847-855) which always returns 0 — so the setup exits 0 with no drop-in on disk, and service accounts are silently unable to log in. Check the write (`[ -s "$drop_in" ]` or `|| return 1` after the heredoc). Same in `ob-backend-setup` (:687).

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+# without the Unix account existing yet. See doc/service-accounts.md.
+AuthorizedKeysCommand $helper %u
+AuthorizedKeysCommandUser nobody
+
+# Required for pam_openbastion to re-check the key's fingerprint against
+# service-accounts.conf: it is what puts SSH_USER_AUTH in the PAM environment.
+# Without it sshd accepts the key and the fingerprint is never verified.
+ExposeAuthInfo yes
+SVCKEYS
+    chmod 644 "$drop_in"
```
guimard: After this write, nothing re-validates sshd: the only `sshd -t` in the script is at :404, BEFORE this function runs. A malformed keyword in the generated file — the PR body itself concedes "a malformed drop-in would break sshd outright" — is caught nowhere until `restart_sshd || warn "Failed to restart SSH"` at :2298. And even that warn never fires: `restart_sshd` is the LEFT operand of `||`, errexit is off in its body, so a failed `systemctl restart sshd` returns 0, prints "Restarted sshd.service" (verified empirically), and setup ends "Configuration Complete" with sshd DOWN. A `sshd -t` right after this chmod closes both holes. Same in `ob-backend-setup` (:702).

doc/service-accounts.md
```diff
@@ -241,9 +241,23 @@ want both:
...
+- `ob-bastion-setup --enable-service-keys` (same flag on `ob-backend-setup`) writes the
+  whole drop-in — `AuthorizedKeysCommand`, `AuthorizedKeysCommandUser` and
+  `ExposeAuthInfo yes` — and creates `service-accounts.d`. This is the flag to use;
+  it exists because #263 showed the manual assembly is easy to get half-right.
+- `--max-security` (Mode E) writes `ExposeAuthInfo yes` on its own, but **not** the
+  `AuthorizedKeysCommand`. Mode E alone therefore still leaves a service account unable
+  to log in.
+
+Neither is on by default: both change sshd for every session on the host, and this
+project's convention is that such changes are opted into. On a host set up without
```
guimard, line 260: "the drop-in above" is the 2-line block at :208-212 — `AuthorizedKeysCommand` + `AuthorizedKeysCommandUser` only, no `ExposeAuthInfo yes` — and this PR removed the sentence that used to tell operators to add it ("On any other host, add it yourself in the drop-in above (#263)"). Following this instruction reproduces the exact half-configuration the PR exists to kill: sshd accepts the key, `SSH_USER_AUTH` is never exposed, and the fingerprint check is silently skipped (as :227-233 explains). The table row at :241 ("`ExposeAuthInfo yes` | the sshd drop-in above") reinforces the mismatch. Add the third directive to the block (or point at the generated file's contents).

doc/service-accounts.md
```diff
@@ -241,9 +241,23 @@ want both:
...
+either, write the drop-in above by hand.
+
+`fingerprint_required = true` is a separate decision and applies to **every** SSH login
```
guimard, line 263: This says the operator can make `fingerprint_required = true` "a separate decision" — i.e. edit openbastion.conf — but the setup scripts silently destroy that manual edit on every run: `render_openbastion_conf` emits no `fingerprint_required` key, and `configure_pam_openbastion` unconditionally does `backup_file "$OB_CONFIG"; render_openbastion_conf > "$OB_CONFIG"` in Phase 1 (with `refresh_openbastion_conf` re-rendering after enrollment) — all BEFORE the new function's own grep (ob-bastion-setup:847) then warns "fingerprint_required is not set". The operator's line is wiped by the very run that reports it missing, and the module silently returns to default-false skip-if-no-fingerprint (src/config.c:157). The doc should say where a manual setting can persist — or the render should pass it through.

tests/test_ob_service_account_keys.sh
```diff
@@ -141,31 +141,44 @@ test_doc_does_not_claim_pam_is_enough() {
...
+                *) bad="$bad $f:$n(in ${fun%%(*}, unguarded)" ;;
+            esac
+        done < <(awk '!/^[[:space:]]*#/ && /ExposeAuthInfo yes/{print NR}' "$src")
+
+        [ "$found" -gt 0 ] || bad="$bad $f(no-write-at-all)"
+
+        # Each of those two functions must early-return when its flag is off.
+        for fun in configure_max_security_sshd configure_service_account_keys; do
+            sed -n "/^$fun() {/,/^}/p" "$src" | head -4 \
+                | grep -qE '(MAX_SECURITY|ENABLE_SERVICE_KEYS)" != "true"' \
```
guimard: This alternation accepts EITHER flag for EITHER function: the loop knows which function it extracted, but the pattern doesn't. Mutation-proven: swapping `configure_max_security_sshd`'s guard to `ENABLE_SERVICE_KEYS" != "true"` keeps the suite green (7/7) — and then `--max-security` alone stops writing the 60- drop-in while print_summary still prints "✓ Maximum Security Mode E", or `--enable-service-keys` alone writes `AuthorizedKeysFile none` + `RevokedKeys` on a host where `download_krl` never ran, and the end-of-setup restart kills sshd. Use the per-function flag name: `'MAX_SECURITY" != "true"'` for the first, `'ENABLE_SERVICE_KEYS" != "true"'` for the second.

tests/test_ob_service_account_keys.sh
```diff
@@ -141,31 +141,44 @@ test_doc_does_not_claim_pam_is_enough() {
...
+            fun=$(awk -v n="$n" 'NR<=n{if(/^[a-z_]+\(\) \{/) x=$0} END{print x}' "$src")
+            case "$fun" in
+                configure_max_security_sshd*|configure_service_account_keys*) ;;
+                # --help text describes the directive; it writes nothing.
+                usage*|show_help*) found=$((found - 1)) ;;
+                *) bad="$bad $f:$n(in ${fun%%(*}, unguarded)" ;;
+            esac
+        done < <(awk '!/^[[:space:]]*#/ && /ExposeAuthInfo yes/{print NR}' "$src")
+
+        [ "$found" -gt 0 ] || bad="$bad $f(no-write-at-all)"
```
guimard: `found` is a union across all writes: it is satisfied by the service-keys write alone, and the guard loop below checks only that both functions have an early return. The rewrite therefore dropped the old test's pin that `configure_max_security_sshd` ITSELF writes `ExposeAuthInfo yes`. Mutation-proven: deleting that line from the 60-max-security.conf heredoc (ob-bastion-setup:879, ob-backend-setup:746) leaves the suite 7/7 green — and then doc/service-accounts.md:250 ("--max-security (Mode E) writes ExposeAuthInfo yes on its own") becomes false with zero failing tests: Mode E hosts lose the fingerprint re-check, the exact half-configuration this suite exists to prevent. Assert per-function presence, not just `found > 0`.

tests/test_ob_service_account_keys.sh
```diff
@@ -141,31 +141,44 @@ test_doc_does_not_claim_pam_is_enough() {
...
-        # would be written on every host, with the enclosing function unchanged
-        # and this test none the wiser.
-        awk '/^configure_max_security_sshd\(\) \{/{f=1} f&&/MAX_SECURITY.*!=.*true/{g=1} f&&/^\}/{exit} END{exit !g}' \
-            "$ROOT_DIR/scripts/$f" || bad="$bad $f(no-early-return-guard)"
+        local src="$ROOT_DIR/scripts/$f" found=0
+        # Non-comment writes only: a mention in prose is not a write.
+        while IFS= read -r n; do
+            [ -n "$n" ] || continue
+            found=$((found + 1))
+            fun=$(awk -v n="$n" 'NR<=n{if(/^[a-z_]+\(\) \{/) x=$0} END{print x}' "$src")
```
guimard: Attribution is by the last function-DEFINITION line at or before the write, with no brace-containment check and a prefix-glob allowlist — so a write that is NOT inside any function, but sits after `configure_service_account_keys`' definition, inherits the allowed name and passes. Mutation-proven: an unconditional top-level `cat > /etc/ssh/sshd_config.d/99-…conf <<EOF` + `ExposeAuthInfo yes` pasted between the two function definitions leaves the suite 7/7 green. Same hole for writes inside `usage_something()` / `show_help_x()` (prefix globs, and `usage*` even decrements `found`), and the guard check (:170-173) asserts only guard TEXT in the first 4 lines — never that each write's line number is AFTER its function's guard line. Assert containment: write line ∈ (function start, function end) and write line > guard line.

scripts/ob-bastion-setup
```diff
@@ -760,6 +761,99 @@ enable_heartbeat_timer() {
...
+        return 0
+    fi
+
+    step "Configuring the SSH layer for service accounts..."
+
+    local helper="/usr/sbin/ob-service-account-keys"
+    local keys_dir="/etc/open-bastion/service-accounts.d"
+    local drop_in="$SSHD_CONFIG_DIR/09-open-bastion-service-keys.conf"
+
+    if [ "$DRY_RUN" = "true" ]; then
```
guimard: The DRY_RUN short-circuit sits BEFORE the two fatal precondition gates (:802 `[ ! -x "$helper" ]` and :808 `[ ! -d "$SSHD_CONFIG_DIR" ]`), so `--dry-run --enable-service-keys` prints "[DRY-RUN] Would write …" and exits 0 on exactly the hosts where a real run aborts mid-Phase-3 (upgraded host without the packaged helper; host without sshd_config.d) — the preflight gives false confidence precisely for the failure modes the flag introduces. The codebase's own pattern does it the other way: `setup_hardening` (:1314-1317) checks the condition FIRST and dry-runs inside it, warning "(would abort in real run)". Same in `ob-backend-setup` (:664).

tests/test_ob_service_account_keys.sh
```diff
@@ -289,12 +302,56 @@ test_keys_dir_is_shipped() {
...
+
+        # It must write ExposeAuthInfo itself. Leaving it to Mode E would ship
+        # an AuthorizedKeysCommand whose fingerprint check silently never runs
+        # -- the exact half-configuration doc/service-accounts.md warns about.
+        sed -n '/^configure_service_account_keys() {/,/^}/p' "$src" \
+            | grep -q 'ExposeAuthInfo yes' || bad="$bad $f(no-ExposeAuthInfo)"
+
+        # And the canonical helper path, not a rediscovered one.
+        sed -n '/^configure_service_account_keys() {/,/^}/p' "$src" \
+            | grep -q "$CANONICAL" || bad="$bad $f(wrong-helper-path)"
```
guimard: `$CANONICAL` here is satisfied by the `local helper="/usr/sbin/ob-service-account-keys"` declaration in the function (ob-bastion-setup:793), not by the heredoc write — and nothing in this suite (or any other: `git grep AuthorizedKeysCommandUser` in tests/ hits only a comment) pins `AuthorizedKeysCommandUser nobody` or the `%u` argument. A regression to `AuthorizedKeysCommandUser root` (helper runs as root) or a dropped `%u` (helper gets no username, serves no key) ships with the suite fully green; even deleting the whole `AuthorizedKeysCommand` line passes. Execution harnesses that would catch this already exist in tests/: `source_script()` (test_ob_bastion_setup.sh:17) and the docker `sshd -T` harness (test_ob_sshd_dropin_precedence.sh).

scripts/ob-bastion-setup
```diff
@@ -1848,6 +1942,13 @@ Options:
                           ($OB_TOKEN), in which case it is reused.
   -k, --insecure          Skip SSL certificate verification
   --max-security          Enable Mode E (SSO certificates only, sudo via LLNG token, KRL)
+  --enable-service-keys   Configure the SSH layer for service accounts: writes
```
guimard, line 2078: The flag is documented here, but it is invisible everywhere an operator can confirm it: print_summary's "Features enabled" block (:1821-1862) has rows for hardening, audit, Mode E and sudo-fresh-otp but none for this feature, and the "reports whether fingerprint_required is set" promise is only the negative 5-line warn in the function (:847-853) — which no test pins (deleting the block keeps the suite green) and which never emits a positive confirmation when the setting IS present. A non-interactive `--yes` run ends "Configuration Complete" with a record byte-identical to a run without the flag — for a change the CHANGELOG itself describes as altering sshd for every session on the host. Add a summary row and pin the report block. Same in `ob-backend-setup` (:1898).

guimard:
**local-test/lib.sh:335-338** (file not modified by this PR, hence a PR-level comment): the project's own lab provisioning still ships the two-directive drop-in — `AuthorizedKeysCommand` + `AuthorizedKeysCommandUser`, no `ExposeAuthInfo yes` — on every non-Mode-E lab VM (deploy-shell.sh:128 → lab-d). That is the exact shape tests/test_ob_service_account_keys.sh:330-332 denounces as "the silent half-configuration", yet lib.sh:362 asserts "fingerprint matched, no SSO" on a pipeline that supplies no `SSH_USER_AUTH` to verify anything. The lab currently green-lights the very defect #263 was reported for. Consider provisioning lab service keys via `--enable-service-keys` (or adding the third line).

guimard:
Les treize points sont corrigés (`f319641`). Chacun vérifié au source avant correction — la revue a raison partout où ça compte.

**Les trois qui auraient cassé un hôte de production**

1. **L'écrasement silencieux d'un `AuthorizedKeysCommand` préexistant** : confirmé, `Include` en ligne 1 (`:372`) et première valeur gagnante — le commentaire du script le dit lui-même. La fonction **refuse** maintenant avant toute écriture, en expliquant comment combiner les deux. Notre propre drop-in est ignoré par la sonde, donc rejouer reste idempotent.

2. **La chaîne d'écriture ne pouvait pas échouer bruyamment** : exact, et c'est le fait errexit que j'avais moi-même établi pour #257. Chaque étape est vérifiée.

3. **Rien ne revalidait sshd** — et le filet supposé n'existe pas : `restart_sshd` termine sur un `info` qui rend 0, donc `restart_sshd || warn` ne peut jamais avertir. `sshd -t` tourne désormais avec le fichier en place, et le retire si sshd le rejette.

**Les deux où je contredisais mes propres affirmations**

4. Le drapeau était à sens unique. Un run sans le drapeau retire le drop-in — mais **seulement s'il porte notre en-tête** ; un fichier écrit à la main est conservé et signalé.

5. `fingerprint_required` était effacé par le run même qui avertissait de son absence. Il est maintenant reporté par `render_openbastion_conf`.

**Sur la doc, tu as mis le doigt sur une régression que j'avais introduite** : le bloc à deux directives avait perdu `ExposeAuthInfo` quand je l'ai réécrit, donc suivre la doc reproduisait la demi-configuration que la PR existe pour tuer. Troisième ligne rétablie. Et `local-test` livrait le même fichier à deux lignes sur chaque VM non-Mode-E — le lab validait le défaut de #263.

**Les quatre défauts de tests**, tous rejoués : la garde acceptait l'un ou l'autre drapeau pour l'une ou l'autre fonction ; `found` était une union qui laissait supprimer `ExposeAuthInfo` du Mode E en restant vert ; l'attribution se faisait par « dernière définition au-dessus » et laissait passer une écriture top-level ; rien n'épinglait `AuthorizedKeysCommandUser` ni `%u`. C'est de la **containment** maintenant — l'écriture doit tomber dans la plage d'accolades d'une fonction **et** après sa garde.

**Trouvé en corrigeant** : trois assertions de la suite étaient `sed … | head -4` / `| grep -q`, ce qui donne un SIGPIPE à `sed` dès que le consommateur sort — `pipefail` en fait un pipeline échoué, et l'assertion signalait un raté sur un corps qui contenait bien la ligne. Ça n'a émergé qu'en grandissant la fonction. Capture d'abord, match ensuite — le même piège qu'en revue de #264.

Vérifié dans un conteneur `debian:trixie` avec un vrai sshd : activation sur hôte propre → drop-in écrit et `sshd -t` OK ; désactivation → le nôtre retiré ; fichier manuel → conservé et signalé ; `AuthorizedKeysCommand` étranger → refus RC=1, **y compris sous `--dry-run`**.

35 suites sans skip, shellcheck inchangé, prettier propre.

Issue #268, opened 2026-09-07
0.7.0 release checklist: two missing tests, three stale documents, one decision


guimard:
Everything below is verified against `main` at the time of writing, and is what
stands between the current tree and a defensible `0.7.0` tag. Ordered by what
changes the security claim, not by effort.

## 1. `find_anchor()` has two implementations and no test

`src/ob-fp-daemon.c:250` and `src/pam_openbastion.c:1774` implement the same
walk, and the daemon's own comment says:

> This MUST agree exactly with pam_openbastion's find_sshd_session_ancestor(),
> or the module looks for a drop under a key this daemon never wrote.

Nothing verifies that sentence: `grep -rl 'find_anchor\|find_sshd_session_ancestor' tests/`
returns nothing. A divergence breaks the SSH fingerprint binding **silently** —
no error at login, the module simply finds no drop — and that binding is what
`doc/security/99-risk-reduce.md` credits with reducing R-S3 and R-S15.

This is the same class of drift that #254 removed for the systemd units, left
in place on the mechanism whose trust root #249 had just moved.

**Acceptance:** one test drives both implementations over the same synthetic
`/proc` ancestry and requires identical output, including the edge cases the
walk defines (depth limit, non-contiguous `sshd-session`, pid 1). It goes in
`tests/mutation/catalogue` afterwards, so a future divergence fails rather than
being noticed by reading.

## 2. `src/cache_key.c` rejection branches have no test

`4cbe85c` added them and touched no test file; there is no `tests/test_cache_key*`.
Six `rejected` paths are uncovered: not a regular file, not root-owned,
setuid/setgid bits, and group/other permission bits.

It is the only hardening of the 0.7.0 series with no proof, and it corrects a
real branch-nesting bug.

**Acceptance:** unit tests for each refusal, and a catalogue entry so removing
one fails.

## 3. Documentation that contradicts the tree

Small, no design decisions, and an evaluator reads these before reading code:

- `CMakeLists.txt:2` still says `VERSION 0.6.2`.
- `doc/security/07-plan-de-traitement.md`: **MT52** is marked `Ouvert` although
  #247 and #252 are merged.
- `SECURITY.md:539` still lists "Encrypted secrets in the secret store become
  permanently unrecoverable" among the impacts of a machine-id change. The main
  passage was corrected when the store was removed; this bullet in the
  machine-id section was not.

## 4. A process note in the homologation dossier

[...]

PR #272, opened 2026-09-08
feat(r-p1): make the portal configuration a release prerequisite, and say so everywhere

guimard:
Item 5 of #268 — the release gate, and the only item that changes what the
product claims.

## The decision

With `pamAccessServerGroups` empty — **the shipped default**, and the
multi-group model `doc/bastion-architecture.md` recommended — `server_group` is
read from the request body, so any enrolled host of the project that is
compromised can declare itself a bastion on `/pam/authorize` and obtain a hop
voucher for a user. `allowed_bastions` empty still means "accept any", for
Ansible compatibility.

#268 named two honest positions. This PR takes the first: **release
prerequisite**, not accepted residual risk. A shipped default that contradicts
the configuration the architecture document recommends is a condition of
service, and it is not recoverable after the fact — a host compromised during
the window already has its vouchers.

## Record

- **CE03** loses its *"ou mode hérité explicitement accepté"* escape and becomes
  blocking; **CE06** and **CE16** are marked blocking.
- **CE21** is new: `pamAccessAllowedRps` + bastions under a `client_id` of their
  own. Without it the upstream audience binding does nothing at all — an empty
  list means "no change".
- **MT40** and **MT50** become publication prerequisites; **MT53** is the
  `pamAccessAllowedRps` row the plan had no entry for.
- The R-P1 sheet says the residual score is no longer conditioned on an
  assumption.

## Three operator docs said the opposite

`doc/llng-configuration.md`, `doc/admin-guide.md` and
`doc/bastion-architecture.md` all told the reader to leave
`pamAccessServerGroups` empty *"for the usual model of one `client_id` per
project covering several server groups"*. That model is precisely the
configuration in which the gap is exploitable. All three now say the opposite —
**including what it costs**: one `client_id` per server group, because an
unmapped one is refused. Saying "this is now required" without saying "and here
is the enrolment work it implies" would only move the surprise.

## Enforcement (what is possible, and what is not)

The product can neither set nor verify the portal half. These are LLNG Manager
settings, and reading them back from a host would need an API that publishes the
project's bastions, server groups and RPs — which is exactly what an SSO must
not offer. So the enforcement is declarative, and it is on every path an
operator takes:

| where | what |
|---|---|
| `ob-bastion-setup`, `ob-backend-setup`, `ob-desktop-setup` | printed in the summary, every run |
| `ob-post-upgrade` | printed — it is how an existing host reaches 0.7.0, the only path to hosts set up before any of this existed |
| `ob-builder` | writes `PORTAL-CHECKLIST.md` next to each artefact, **pre-filled** with that deployment's `client_id` and `server_group` |
| `UPGRADE-NOTES.md` | new **B0**; Part B stops being "only if you are also moving the portal" |

The builder checklist matters most for the Ansible role, whose setup task runs
with nobody reading its stdout.

The text is shipped **once**, at
`/usr/lib/open-bastion/ob-portal-prerequisites.txt`, and packaged in the `.deb`
and the `.rpm` as well as by cmake — a helper that existed in the tree and not
in the package is #263, and a warning nobody receives is worse than none.

## Tests

`tests/test_ob_portal_prerequisites.sh` runs the code rather than reading it:
it executes each script's printer, runs `ob-post-upgrade --dry-run`, and drives
`ob-builder`'s `run_outputs_for_role` with the renderers stubbed to check both
artefact kinds get a checklist with no unresolved `@@PLACEHOLDER@@`.

```
$ bash tests/test_ob_mutation.sh
  PASS: portal-prereq-bastion-setup — tests/test_ob_portal_prerequisites.sh catches it
  PASS: portal-prereq-post-upgrade — tests/test_ob_portal_prerequisites.sh catches it
  PASS: portal-prereq-builder-ansible — tests/test_ob_portal_prerequisites.sh catches it
Tests run: 14, passed: 14, failed: 0
```

Every `tests/test_ob_*.sh` suite green, `ctest` 22/22.

## Also

The dossier's **CE02** row was repaired: unescaped pipes inside the
`^/ssh/(admin|certs|revoke)` regex had collapsed it into the neighbouring
columns, taking R-S4 and R-P2 with it. It is the table this PR adds a row to.


orontee:
Is it expected that the CI job "Build Debian Package" fails silently? See https://github.com/linagora/open-bastion/actions/runs/34227502715/job/102065857202
```
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: open-bastion
CPack: - Install project: open-bastion []
CMake Error: failed to create symbolic link '/home/runner/work/open-bastion/open-bastion/build/_CPack_Packages/Linux/DEB/open-bastion-0.6.2-Linux/usr/local/sbin/ob-standalone-setup': No such file or directory
CMake Error: failed to create symbolic link '/home/runner/work/open-bastion/open-bastion/build/_CPack_Packages/Linux/DEB/open-bastion-0.6.2-Linux/usr/local/share/man/man8/ob-standalone-setup.8': No such file or directory
CPack: Create package
CPack: - package: /home/runner/work/open-bastion/open-bastion/build/open-bastion-0.6.2-Linux.deb generated.
```

Note that `cd build; cpack -G DEB` fails on my workstation too with same error.

guimard:
@orontee Good catch, thanks — no, it is not expected, and it is not specific to this PR: main's CI shows the same two lines.

The cpack `.deb` really ships without `ob-standalone-setup` and its man page. The two links were created by `install(CODE)` against `CMAKE_INSTALL_FULL_SBINDIR` / `CMAKE_INSTALL_FULL_MANDIR`, which are fixed at configure time (`/usr/local/...`), while cpack installs under `/usr`: the parent directory does not exist in the staging tree, `create_symlink` fails, `execute_process` ignores it and cpack still exits 0.

The released packages are not affected: `dpkg-buildpackage` and `rpmbuild` configure with prefix `/usr`, so the paths match there (and both list the link explicitly).

Fix in #273: the links are now installed with `install(FILES)` and a relative destination, and the CI job fails when cpack reports a `CMake Error`.

orontee:
@guimard I confirm that #273 fixed the package build.

guimard:
> @orontee Good catch, thanks — no, it is not expected, and it is not specific to this PR: main's CI shows the same two lines.
> 
> The cpack `.deb` really ships without `ob-standalone-setup` and its man page. The two links were created by `install(CODE)` against `CMAKE_INSTALL_FULL_SBINDIR` / `CMAKE_INSTALL_FULL_MANDIR`, which are fixed at configure time (`/usr/local/...`), while cpack installs under `/usr`: the parent directory does not exist in the staging tree, `create_symlink` fails, `execute_process` ignores it and cpack still exits 0.
> 
> The released packages are not affected: `dpkg-buildpackage` and `rpmbuild` configure with prefix `/usr`, so the paths match there (and both list the link explicitly).
> 
> Fix in #273: the links are now installed with `install(FILES)` and a relative destination, and the CI job fails when cpack reports a `CMake Error`.

Let's merge #272 then ?

orontee:
It's ok for me.

Issue #279, opened 2026-09-18
As security certification manager, I want to publish a complete and crystal-clear documentation


orontee:
# Context

For certification one expects:
- An end-user manual
- An administrator manual
- A user guide

We must be able to share their content with non technical users and provide them ways to navigate the documentation. A clear view of the version one is reading is also important.

The current documentation has valuable content but lacks organization and suffer repetitions.

# Suggestion

* [x] Use [Sphinx](https://www.sphinx-doc.org/en/master/index.html) to be able to transform the Markdown sources in structured HTML or PDF with index and summary. It will also help maintaining cross references.

* [ ] Rework the organization of the documentation.

* [x] Switch from Markdown to ReST.

PR #280, opened 2026-09-21
docs: convert the two document sets to reStructuredText, build them with Sphinx

guimard:
Converts the two document sets under `doc/` to reStructuredText, builds them as one Sphinx project, gates that build in CI, and ships the HTML as a separate Debian package.

## The two docs, one project

| Part | Was | Is |
| --- | --- | --- |
| English technical documentation | `doc/*.md` + `doc/design/` | `doc/*.rst`, entry point `doc/index.rst` |
| French EBIOS Risk Manager study | `doc/security/*.md` | `doc/security/*.rst`, entry point `doc/security/index.rst` |

One `conf.py`, one HTML site, two parts in the sidebar. The two `README.md` files became the two `index.rst`: they keep the curated tables they always had, and carry hidden toctrees so navigation works without a generated list duplicating the tables above it.

## Why reST is worth the diff

A cross-document link is now a `:doc:`/`:ref:` role resolved at build time. A pointer to a page or a section that no longer exists fails the build instead of 404-ing for the reader — which is the whole reason the conversion rewrote every internal link instead of leaving the Markdown ones alone:

- 247 `:doc:` roles, 54 anchored `:ref:`s with explicit labels on the target sections;
- links out of the Sphinx tree (`quick-start/`, `admin-builder/`, the man pages) became absolute GitHub URLs;
- where the link text was just a file name, it is gone — Sphinx substitutes the target's title, which cannot go stale;
- one anchor in `audit.md` was already truncated and broken; it resolves now.

Mermaid diagrams keep their directive: `conf.py` uses `sphinxcontrib-mermaid` when installed and falls back to a literal block when not, so the build needs nothing beyond Sphinx itself.

## CI

A `docs` job builds `doc/` with `sphinx-build -W --keep-going`: a dead `:doc:`/`:ref:` target, a document in no toctree and a table docutils cannot parse all fail there. It then builds the same tree through CMake — the path the package is filled from — and uploads the HTML as an artefact. It also refuses a `.md` file under `doc/`: one dropped back in there would be in no toctree, would not be built and would not be packaged.

The packaging build deliberately does **not** pass `-W`. A warning is a documentation defect, and the place to fail on one is review, not a release build whose Sphinx version happens to word something differently.

## Packaging

`-DBUILD_DOC=ON` (off by default, so an ordinary build still needs no Python) builds the HTML; `debian/rules` turns it on and the result goes to a new arch-all **`open-bastion-doc`**, 287 kB, installable on its own. `dh --with sphinxdoc` deduplicates the JavaScript against `libjs-sphinxdoc`. The doctree cache, `.buildinfo` and the empty `_sources/` are kept out; `open-bastion-doc` is lintian-clean.

## Tests touched

`tests/ebios_matrix_check.py` reads the EBIOS study, so it moved with it: reST sections instead of ATX headings, and a grid-table reader in place of the pipe-row regexes — a matrix cell spanning two physical lines would otherwise come back holding half its risks. It still parses **47 sheets**, and was re-checked against three deliberate mutations (a moved matrix cell, a changed sheet score, a risk detached from its feared event); each is still caught. `tests/test_ob_changelog.sh` now checks `.rst` link targets too, which its `.md`-only regex would have skipped.

## Verified locally

- `sphinx-build -W --keep-going` clean, both with the RTD theme + mermaid and with Sphinx alone (alabaster fallback);
- `cmake -DBUILD_DOC=ON && cmake --build . --target doc`, then install;
- `dpkg-buildpackage -b`: four binary packages, `open-bastion-doc` lintian-clean;
- all 39 `tests/test_ob_*.sh` suites and `ctest` (22/22) pass.

## Note for review

The CHANGELOG's historical entries keep their prose (they describe what was true then) but their **links** are retargeted — a pointer to a file that is gone is worse than the duplication it replaced, and `test_ob_changelog.sh` enforces that they resolve.


guimard:
Follow-up pushed in `995b747`: the release path was broken and I had missed it.

`release.yml` builds its `.deb` with `dpkg-buildpackage`, which checks `Build-Depends` before it starts. This PR added `python3-sphinx` and `python3-sphinx-rtd-theme` to `debian/control` but not to that job's `apt-get install` list, so **the next tag would have failed on all three codenames** — not produced a doc package: failed outright. Fixed.

Everything downstream was already generic, so nothing else needed changing:

| Stage | How `open-bastion-doc` gets there |
| --- | --- |
| `build-deb` (bookworm, trixie, noble) | `mv ../*.deb packages/` |
| `create-repo` | copied into each suite, then `dpkg-scanpackages` + signed `Release`/`InRelease` |
| `deploy` | GitHub Pages APT repo — `apt install open-bastion-doc` |
| `github-release` | attached to the tag with the other packages |

The repository index page now lists `open-bastion-doc` (and `open-bastion-builder`, equally undiscoverable until now) under the install instructions.

Also closed the coverage gap that let this through: nothing in CI went through `debian/rules` — `package-deb` uses cpack. The `docs` job now runs `mk-build-deps` from `debian/control` and `dpkg-buildpackage -b`, then asserts the built `open-bastion-doc_*_all.deb` contains `html/index.html` and at least 30 pages (locally: 39). A missing build-dependency now fails in review, not at the tag.

orontee:
Relates to #279

orontee:
I checked that the `doc` target build successfully locally, with and without the `sphinxcontrib-mermaid` extension. Output looks good, all content seems to be here.

LGTM.

